PR #10, opened 2026-07-06
feat(robotiq_tsf): tactile sensor RViz visualization

ebarnett3:
## Summary

Adds RViz visualization for the TSF-85 tactile sensors: `tactile_viz_node` (Python) turns the raw `TactileSensor/StaticData` stream into per-finger heatmaps and 3D markers, with two ready-to-use launches and RViz configs.

- **`tactile_viz_node`** (`robotiq_tsf/scripts/`): subscribes to `StaticData`, publishes
  - `visualization_msgs/MarkerArray` on `/tactile/markers` (per-taxel 3D cells on each finger pad),
  - two `sensor_msgs/Image` heatmaps (one per finger) on `/tactile_viz/finger{1,2}_heatmap`,
  - the pad TFs (optional — `publish_static_tfs`).

  On startup it averages the first `baseline_frames` messages into a per-taxel baseline and subtracts it (raw taxels carry a large DC offset per cell); re-zero anytime via `std_msgs/Empty` on `/tactile_viz/zero`. Residuals below `noise_floor` render quiet. Pad geometry, frames, color scale and heatmap options are all parameters.
- **`tactile_viz.launch.py`**: driver + viz + RViz, minimal standalone view.
- **`gripper_tactile_viz.launch.py`**: single-window combined view — 2F-85 gripper bringup (`robotiq_description`) with the tactile pads TF-mounted on the fingertip links, so the heatmaps follow the fingers as the gripper opens/closes. Delays the viz start (`tactile_delay`, default 8 s) so the baseline is captured after gripper activation, with the pads clear.
- Install rules for `scripts/`, `launch/`, `rviz/`; `package.xml` exec deps; `.gitignore` for `__pycache__`/`*.pyc`.
- README: "Tactile visualization (RViz)" section under TSF-85 — launches, common args, topics, baseline zeroing/re-zero.

Both launches drive the existing `poll_data_node`. A `poller:=` launch arg (default `poll_data_node`) lets you swap the driver executable without touching the launch files.

## Note on data quality

Depending on the USB enumeration state (it can toggle with a replug), `poll_data_node` intermittently emits corrupted `StaticData` on the TSF-85 (rail-to-rail 0..65535, per-taxel std in the thousands at rest), which renders as saturated/flashing heatmaps. That is a driver parsing issue, addressed separately by the SDK-backed poller PR — it does not block this PR, which is the visualization tooling. In a healthy enumeration state the heatmaps are clean (verified on hardware with `poll_data_node`, per-taxel std ~3 counts).

## Testing

- 14 unit tests (ament pytest, `test/test_tactile_viz.py`, runs in CI via `colcon test`): baseline capture/re-zero, noise floor and zero-clamping, marker geometry/ids/colors, finger-1 column flip, heatmap image shape/encoding, autoscale, publish rate limiting, malformed-input rejection, colormap and rpy→quat.
- `docker build` passes; verified the image installs the node, launches and rviz configs, and both launches parse (`--show-args`).
- Real hardware: RViz brings up the heatmap displays against `poll_data_node`; baseline zeroing and noise floor behave as intended.

## Screencasts

1.  `./run.sh sensor` and `ros2 launch robotiq_tsf tactile_viz.launch.py` in the container :

[Screencast from 2026-07-05 23-03-09.webm](https://github.com/user-attachments/assets/ccbb1e50-8702-40dc-8f2e-673c27ea3832)

2. `./run.sh both` and `ros2 launch robotiq_tsf gripper_tactile_viz.launch.py com_port:=/dev/ttyUSB0` in the container :

[Screencast from 2026-07-05 23-19-48.webm](https://github.com/user-attachments/assets/c9664c42-e305-428d-9c27-1f8347414127)

Note that the red taxel flashing in the first video is an intermittent problem that I see quite often. It is resolved by using the PollDataSdk technique that we used in the RUC demo. I will propose this in a follow-up MR. Has anyone else encountered this problem?

FYI @l-tschreiber-a11y

### DIFF
--- a/robotiq_tsf/launch/gripper_tactile_viz.launch.py
+++ b/robotiq_tsf/launch/gripper_tactile_viz.launch.py
@@ -22,12 +22,16 @@ from launch.actions import (
     DeclareLaunchArgument, IncludeLaunchDescription, TimerAction,
 )
 from launch.launch_description_sources import PythonLaunchDescriptionSource
-from launch.substitutions import LaunchConfiguration, PathJoinSubstitution
+from launch.substitutions import (
+    LaunchConfiguration, PathJoinSubstitution, PythonExpression,
+)
 from launch_ros.actions import Node
 from launch_ros.substitutions import FindPackageShare
 
 # Tactile pad mount relative to its fingertip link, mirrored left/right:
 # lateral offset toward the pad surface and height along the finger.
+# Defaults for the 2F-85; override the mount_* / *_tip_link launch args to
+# remount the pads or use another gripper (e.g. 2F-140).
 TACTILE_MOUNT_X_M = 0.0252
 TACTILE_MOUNT_Z_M = 0.032
 
@@ -45,12 +49,22 @@ def generate_launch_description():
             'com_port', default_value='/dev/ttyUSB0',
             description='Gripper serial port.'),
         DeclareLaunchArgument(
-            'poller', default_value='poll_data_node',
-            description='Driver executable publishing StaticData. Override to '
-                        'point the viz at an alternative poller.'),
-        DeclareLaunchArgument(
             'use_fake_hardware', default_value='false',
             description='Gripper ros2_control mock instead of real hardware.'),
+        DeclareLaunchArgument(
+            'mount_x', default_value=str(TACTILE_MOUNT_X_M),
+            description='Pad lateral offset from its fingertip link [m] '
+                        '(negated for the left finger).'),
+        DeclareLaunchArgument(
+            'mount_z', default_value=str(TACTILE_MOUNT_Z_M),
+            description='Pad height along the finger from its fingertip '
+                        'link [m].'),
+        DeclareLaunchArgument(
+            'left_tip_link', default_value='robotiq_85_left_finger_tip_link',
+            description='Parent link for the finger-0 pad frame.'),
+        DeclareLaunchArgument(
+            'right_tip_link', default_value='robotiq_85_right_finger_tip_link',
+            description='Parent link for the finger-1 pad frame.'),
         DeclareLaunchArgument(
             'rviz_config', default_value=default_rviz,
             description='Combined gripper + tactile RViz config.'),
@@ -81,23 +95,25 @@ def generate_launch_description():
     anchor_0 = Node(
         package='tf2_ros', executable='static_transform_publisher',
         name='finger_tip_to_tactile_finger_0',
-        arguments=['--x', str(-TACTILE_MOUNT_X_M), '--y', '0',
-                   '--z', str(TACTILE_MOUNT_Z_M),
-                   '--frame-id', 'robotiq_85_left_finger_tip_link',
+        arguments=['--x', PythonExpression(['-', LaunchConfiguration('mount_x')]),
+                   '--y', '0',
+                   '--z', LaunchConfiguration('mount_z'),
+                   '--frame-id', LaunchConfiguration('left_tip_link'),
                    '--child-frame-id', 'tactile_finger_0'])
     anchor_1 = Node(
         package='tf2_ros', executable='static_transform_publisher',
         name='finger_tip_to_tactile_finger_1',
-        arguments=['--x', str(TACTILE_MOUNT_X_M), '--y', '0',
-                   '--z', str(TACTILE_MOUNT_Z_M),
+        arguments=['--x', LaunchConfiguration('mount_x'),
+                   '--y', '0',
+                   '--z', LaunchConfiguration('mount_z'),
                    '--yaw', str(math.pi),
-                   '--frame-id', 'robotiq_85_right_finger_tip_link',
+                   '--frame-id', LaunchConfiguration('right_tip_link'),
                    '--child-frame-id', 'tactile_finger_1'])
 
-    # Sensor driver — autodetects the device.
-    poll = Node(
-        package='robotiq_tsf', executable=LaunchConfiguration('poller'),
-        name='poll_data', output='screen')
+    # Sensor driver (poller:= arg declared there).
+    driver = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource(PathJoinSubstitution([
+            FindPackageShare('robotiq_tsf'), 'launch', 'tsf_driver.launch.py'])))
 
     # publish_static_tfs:=false — render into the gripper-mounted tactile_finger_*
     # frames (from the anchors above) instead of the node's standalone frames.
@@ -122,4 +138,4 @@ def generate_launch_description():
         period=LaunchConfiguration('tactile_delay'), actions=[viz])
 
     return LaunchDescription(
-        args + [gripper, anchor_0, anchor_1, poll, viz_delayed, rviz])
+        args + [gripper, anchor_0, anchor_1, driver, viz_delayed, rviz])
--- a/robotiq_tsf/launch/gripper_tactile_viz.launch.py
+++ b/robotiq_tsf/launch/gripper_tactile_viz.launch.py
@@ -1,0 +1,125 @@
+"""Single-window view: the 2F-85 gripper with the tactile heatmaps mounted on
+its fingers.
+
+Brings up, in one RViz:
+  - the gripper (robotiq_control.launch.py: ros2_control + robot_state_publisher),
+  - two static TFs mounting the tactile pad frames on the gripper fingertip
+    links (values from the demo), so the heatmaps move with the fingers,
+  - the tactile driver + tactile_viz_node (publish_static_tfs:=false, so it
+    renders into the gripper-mounted frames rather than its own),
+  - RViz showing the RobotModel + MarkerArray + per-finger Image heatmaps.
+
+  ros2 launch robotiq_tsf gripper_tactile_viz.launch.py com_port:=/dev/ttyUSB0
+
+Run this INSTEAD of the separate gripper / tactile_viz launches.
+"""
+import math
+import os
+
+from ament_index_python.packages import get_package_share_directory
+from launch import LaunchDescription
+from launch.actions import (
+    DeclareLaunchArgument, IncludeLaunchDescription, TimerAction,
+)
+from launch.launch_description_sources import PythonLaunchDescriptionSource
+from launch.substitutions import LaunchConfiguration, PathJoinSubstitution
+from launch_ros.actions import Node
+from launch_ros.substitutions import FindPackageShare
+
+# Tactile pad mount relative to its fingertip link, mirrored left/right:
+# lateral offset toward the pad surface and height along the finger.
+TACTILE_MOUNT_X_M = 0.0252
+TACTILE_MOUNT_Z_M = 0.032
+
+
+def generate_launch_description():
+    pkg_tsf = get_package_share_directory('robotiq_tsf')
+    default_rviz = os.path.join(pkg_tsf, 'rviz', 'gripper_tactile.rviz')
+
+    com_port = LaunchConfiguration('com_port')
+    use_fake_hardware = LaunchConfiguration('use_fake_hardware')
+    rviz_config = LaunchConfiguration('rviz_config')
+
+    args = [
+        DeclareLaunchArgument(
+            'com_port', default_value='/dev/ttyUSB0',
+            description='Gripper serial port.'),
+        DeclareLaunchArgument(
+            'poller', default_value='poll_data_node',
+            description='Driver executable publishing StaticData. Override to '
+                        'point the viz at an alternative poller.'),
+        DeclareLaunchArgument(
+            'use_fake_hardware', default_value='false',
+            description='Gripper ros2_control mock instead of real hardware.'),
+        DeclareLaunchArgument(
+            'rviz_config', default_value=default_rviz,
+            description='Combined gripper + tactile RViz config.'),
+        DeclareLaunchArgument(
+            'tactile_delay', default_value='8.0',
+            description='Seconds to wait before starting tactile_viz (and thus '
+                        'its baseline capture), so the gripper finishes its '
+                        'activation open/close and settles OPEN — pads clear — '
+                        'first. The driver streams from t=0 regardless.'),
+    ]
+
+    # Gripper: ros2_control + robot_state_publisher (its own RViz suppressed —
+    # we run the combined one below).
+    gripper = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource(PathJoinSubstitution([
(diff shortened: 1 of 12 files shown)


jkski-robotiq:
for the flashing grid, i think this is constent with processes not being shut down correctly and multple nodes accessing the tactiel data. THe PollDataSDK was made when I was tryting to debug/troubleshoot that issue, so i think that node checks and shuts down other tactile processes whenere the polldata node does not

robotiq_tsf/launch/gripper_tactile_viz.launch.py
```diff
@@ -0,0 +1,125 @@
...
+        arguments=['--x', str(TACTILE_MOUNT_X_M), '--y', '0',
+                   '--z', str(TACTILE_MOUNT_Z_M),
+                   '--yaw', str(math.pi),
+                   '--frame-id', 'robotiq_85_right_finger_tip_link',
+                   '--child-frame-id', 'tactile_finger_1'])
+
+    # Sensor driver — autodetects the device.
+    poll = Node(
+        package='robotiq_tsf', executable=LaunchConfiguration('poller'),
+        name='poll_data', output='screen')
```
mbegin-robotiq: This driver block (and the `poller` launch arg with its default/description) is duplicated verbatim in `tactile_viz.launch.py`. Suggestion: factor it into a small driver-only launch file that both include, e.g. `launch/tsf_driver.launch.py`:

```python
"""Bring up the TSF-85 driver (StaticData publisher) only."""
from launch import LaunchDescription
from launch.actions import DeclareLaunchArgument
from launch.substitutions import LaunchConfiguration
from launch_ros.actions import Node


def generate_launch_description():
    poller_arg = DeclareLaunchArgument(
        'poller', default_value='poll_data_node',
        description='Driver executable publishing StaticData. Override to '
                    'point the viz at an alternative poller.')

    poll = Node(
        package='robotiq_tsf', executable=LaunchConfiguration('poller'),
        name='poll_data', output='screen')

    return LaunchDescription([poller_arg, poll])
```

Then both launches replace their `poller` arg + `poll` node with:

```python
driver = IncludeLaunchDescription(
    PythonLaunchDescriptionSource(PathJoinSubstitution([
        FindPackageShare('robotiq_tsf'), 'launch', 'tsf_driver.launch.py'])),
    launch_arguments={'poller': LaunchConfiguration('poller')}.items(),
)
```

(Each parent keeps a one-line `DeclareLaunchArgument('poller', default_value='poll_data_node')` so the arg still shows in `--show-args`; the default/description live in one place.)

Side benefit: `tsf_driver.launch.py` becomes a useful standalone bring-up of just the driver, no viz. Since the viz node's config differs between the two launches (`publish_static_tfs`, frames, TimerAction delay), sharing only the driver keeps both files simple — no need to parameterize the viz.

robotiq_tsf/package.xml
```diff
@@ -32,7 +32,21 @@
...
+  <exec_depend>rclpy</exec_depend>
+  <exec_depend>geometry_msgs</exec_depend>
+  <exec_depend>visualization_msgs</exec_depend>
+  <exec_depend>tf2_ros</exec_depend>
+  <exec_depend>python3-numpy</exec_depend>
+  <exec_depend>launch</exec_depend>
+  <exec_depend>launch_ros</exec_depend>
+  <exec_depend>rviz2</exec_depend>
+  <!-- gripper_tactile_viz.launch.py includes the gripper bringup. -->
+  <exec_depend>robotiq_description</exec_depend>
```
mbegin-robotiq, line 45: The node also imports `sensor_msgs.msg.Image` and `std_msgs.msg.{ColorRGBA, Empty}`, but neither `sensor_msgs` nor `std_msgs` is in this new exec_depend block. Are they already declared as `<depend>` higher up for the C++ driver? If not, they should be added here.

robotiq_tsf/scripts/tactile_viz_node
```diff
@@ -0,0 +1,364 @@
...
+        self.declare_parameter('finger_1_rpy', [0.0, 0.0, math.pi])
+        self.declare_parameter('tip_xyz', [0.0, 0.0, 0.0209])
+        self.declare_parameter('tip_rpy', [0.0, 0.0, 0.0])
+        self.declare_parameter('vmin', 0.0)
+        self.declare_parameter('vmax', 700.0)
+        # Set true only if one pad is physically wired with reversed columns.
+        self.declare_parameter('flip_finger_1', False)
+        self.declare_parameter('publish_2d_heatmaps', True)
+        self.declare_parameter('heatmap_topic_finger_0', '/tactile_viz/finger1_heatmap')
+        self.declare_parameter('heatmap_topic_finger_1', '/tactile_viz/finger2_heatmap')
```
mbegin-robotiq: Indexing convention mixes 0- and 1-based: the parameter `heatmap_topic_finger_0` defaults to `/tactile_viz/finger1_heatmap`, while the frames are `tactile_finger_0/1` and the RViz displays say "Finger 1/2". Anyone correlating a heatmap to a marker pad has to remember that finger_0 ≡ Finger 1.

Suggest picking one convention — probably 0-indexed everywhere, matching `msg.taxels[0/1]` — before the topic names ossify into downstream tooling (rviz configs, rosbags, scripts). Cheap to rename now, painful later.

robotiq_tsf/scripts/tactile_viz_node
```diff
@@ -0,0 +1,364 @@
...
+        # without subtraction every cell saturates. Capture a baseline over the
+        # first N frames (keep the sensor untouched at startup) and subtract it;
+        # re-zero any time by publishing std_msgs/Empty on ``zero_topic``.
+        self.declare_parameter('zero_on_start', True)
+        self.declare_parameter('baseline_frames', 500)
+        self.declare_parameter('zero_topic', '/tactile_viz/zero')
+        # Residuals below this (post-baseline) render as zero, so sensor noise
+        # doesn't flash the display. Mirrors the demo's noise floor. Raise if
+        # noise still shows; lower for sensitivity to light touches.
+        self.declare_parameter('noise_floor', 50.0)
```
mbegin-robotiq, line 119: There are two overlapping floor parameters with the same default: `noise_floor` (50.0, post-baseline, applied to both markers and heatmaps in `_on_data`) and `heatmap_floor` (50.0, heatmap-only, applied again in `_grid_to_image`). With both active the heatmap data is floored twice, and a user tuning touch sensitivity has to discover the second knob.

Suggest either defaulting `heatmap_floor` to 0.0 (since `noise_floor` already covers the common case) or adding a comment explaining when you'd want them to differ.

robotiq_tsf/scripts/tactile_viz_node
```diff
@@ -0,0 +1,364 @@
...
+        self._size = float(p('taxel_size_m').value)
+        self._poses = [
+            (list(p('finger_0_xyz').value), list(p('finger_0_rpy').value)),
+            (list(p('finger_1_xyz').value), list(p('finger_1_rpy').value)),
+        ]
+        self._tip_pose = (list(p('tip_xyz').value), list(p('tip_rpy').value))
+        self._vmin = float(p('vmin').value)
+        self._vmax = float(p('vmax').value)
+        if self._vmax <= self._vmin:
+            self._vmax = self._vmin + 1.0
```
mbegin-robotiq, line 141: This silently rewrites a bad `vmax`/`vmin` combination. Worth a `self.get_logger().warn(...)` here — someone who mistypes `vmax:=7.0` as `vmax:=-7.0` will otherwise just see a washed-out display with no hint why.

robotiq_tsf/scripts/tactile_viz_node
```diff
@@ -0,0 +1,364 @@
...
+                m.ns = f'tactile_finger_{fi}'
+                m.id = idx
+                m.type = Marker.CUBE
+                m.action = Marker.ADD
+                m.pose.position = Point(
+                    x=0.0, y=y_half - c*self._pitch_y, z=z_half - r*self._pitch_z)
+                m.pose.orientation.w = 1.0
+                m.scale = Vector3(x=self._size, y=self._size, z=self._size)
+                m.color = ColorRGBA(r=rr, g=gg, b=bb, a=1.0)
+                m.lifetime = Duration(sec=0, nanosec=0)
```
mbegin-robotiq, line 308: Consider `m.frame_locked = True` here. The markers are stamped with time zero ("latest transform"), which works, but in the gripper launch the pad frames move with the fingers — between marker updates the cubes lag the TF by up to one publish interval. `frame_locked` makes RViz re-transform them every render frame, so they track the fingers smoothly. One-line change.

robotiq_tsf/launch/gripper_tactile_viz.launch.py
```diff
@@ -0,0 +1,125 @@
...
+)
+from launch.launch_description_sources import PythonLaunchDescriptionSource
+from launch.substitutions import LaunchConfiguration, PathJoinSubstitution
+from launch_ros.actions import Node
+from launch_ros.substitutions import FindPackageShare
+
+# Tactile pad mount relative to its fingertip link, mirrored left/right:
+# lateral offset toward the pad surface and height along the finger.
+TACTILE_MOUNT_X_M = 0.0252
+TACTILE_MOUNT_Z_M = 0.032
```
mbegin-robotiq, line 36: Optional: these mount offsets (and the `robotiq_85_*_finger_tip_link` frame names below) are hardcoded "values from the demo". If someone remounts the pads or uses the 2F-140, they have to edit the launch file. Exposing them as launch arguments with these defaults would keep the file usable as-is while making it overridable. Fine to defer if this is considered demo-specific for now.

robotiq_tsf/test/test_tactile_viz.py
```diff
@@ -0,0 +1,272 @@
+"""Unit tests for scripts/tactile_viz_node (baseline zeroing, noise floor,
+marker/heatmap generation, colormap, rpy->quat).
+
+The script has no .py extension, so it is loaded via importlib; the node's
+callbacks are driven directly with constructed StaticData messages and the
+publishers are replaced with recording stubs — no executor or driver needed.
```
mbegin-robotiq, line 6: Nice coverage of the numeric behaviors (baseline math, clamping, flip, autoscale). Two gaps, nice-to-have rather than blocking:

- the `publish_static_tfs` path (`_publish_static_tfs`) is never exercised — a stubbed broadcaster could assert the three transforms' frames and poses;
- heatmap pixel content is only checked at pixel (0,0) — a ramp input asserting a couple of interior pixels would pin down row/column orientation of the image, which is exactly the kind of thing that regresses silently.

mbegin-robotiq (commented):
A few small follow-ups from reading through the node, deps, and tests — all minor, none blocking.

mbegin-robotiq (approved):

ebarnett3:
> for the flashing grid, i think this is constent with processes not being shut down correctly and multple nodes accessing the tactiel data. THe PollDataSDK was made when I was tryting to debug/troubleshoot that issue, so i think that node checks and shuts down other tactile processes whenere the polldata node does not

I investigated this problem and it seems that the root cause is a CDC-ACM tty  that is not reset correctly for a fresh enumeration (e.g. when unplugging/replugging the USB), causing packet desync. PollDataSDK uses libserialport for this, which uses raw tty mode and performs the desired reset, thereby fixing the bug. If the same reset is performed manually with the old polldata code, the bug is also resolved: https://github.com/robotiq/ros/pull/11

Note that I'm not sure if it's worth reviewing https://github.com/robotiq/ros/pull/11 and merging it, given that we will probably migrate to PollDataSDK anyway. But it is interesting to use as a reference that isolates the root cause of the problem.

robotiq_tsf/launch/gripper_tactile_viz.launch.py
ebarnett3: Done in 7e4ae18 — factored into `launch/tsf_driver.launch.py` as suggested, included by both viz launches. `poller:=` is declared there and still shows up in `--show-args` for both.

robotiq_tsf/package.xml
ebarnett3: Both are already declared as `<exec_depend>` at the top of the manifest (lines 27–28, alongside `rclcpp`/`ament_index_cpp` for the C++ nodes) — `exec_depend` isn't scoped per-node, so the Python node's imports are covered by those existing entries. The new block only adds the deps that weren't already declared.

robotiq_tsf/scripts/tactile_viz_node
ebarnett3: Done in 7e4ae18 — heatmap topics now default to `/tactile_viz/finger0_heatmap` / `finger1_heatmap`, and the RViz display names and README were updated to "Finger 0/1", matching the frames and `msg.taxels` indexing. Good call doing this before anything downstream depends on the old names.

robotiq_tsf/scripts/tactile_viz_node
ebarnett3: Done in 7e4ae18 — `heatmap_floor` now defaults to 0.0, with a comment scoping it as an extra heatmap-only cutoff on top of `noise_floor` for when the heatmaps specifically need a harsher floor than the 3D view.

robotiq_tsf/scripts/tactile_viz_node
ebarnett3: Done in 7e4ae18 — logs a warning with the rejected value and the fallback before clamping.

robotiq_tsf/scripts/tactile_viz_node
ebarnett3: Done in 7e4ae18 — set on all markers, with a comment on why (pad frames move with the fingers between updates).

robotiq_tsf/launch/gripper_tactile_viz.launch.py
ebarnett3: Done in 7e4ae18 — exposed as `mount_x` / `mount_z` / `left_tip_link` / `right_tip_link` launch args with the 2F-85 values as defaults, so a remount or 2F-140 works without editing the file.

robotiq_tsf/test/test_tactile_viz.py
ebarnett3: Both added in 7e4ae18 — `test_publish_static_tfs` asserts the three transforms' parent/child frames and poses (including the finger-1 yaw-π quaternion) via a stubbed broadcaster, and `test_heatmap_pixel_orientation` feeds a unique-value ramp and asserts six pixels including corners and interior cells, pinning the image's row/column orientation.

PR #11, opened 2026-07-07
fix(robotiq_tsf): force raw tty mode in poll_data_node

ebarnett3:
## Problem

After unplugging/replugging the sensor's USB-C, `poll_data_node` publishes
rail-to-rail garbage on `TactileSensor/StaticData` — the taxels "flash red" in
RViz — and it stays broken across restarts until something else resets the port.
Plain start/stop restarts don't trigger it; a fresh USB re-enumeration does.

## Root cause

`OpenAndConfigurePort()` read the port's existing attributes with `tcgetattr`
and only adjusted `c_cflag` and `ICANON`. It **never reset `c_iflag`**.

A freshly enumerated CDC-ACM tty comes up in *cooked* line discipline, where
`c_iflag` has `ICRNL` (translate `0x0D`→`0x0A`) and `IXON` (consume `0x11`/`0x13`
as XON/XOFF flow control). Those byte values occur constantly in the binary
sensor stream, so the incoming bytes were silently rewritten/dropped, the packet
framing desynced, and every taxel decoded as noise. Because the node wrote the
cooked flags straight back, the corruption persisted across restarts.

termios settings persist on a tty across opens, which is why the failure looked
intermittent: once any raw-mode reader had touched the port, `poll_data_node`
inherited raw mode and worked — until the next replug reset it to cooked.

This was verified on hardware: on a fresh replug the tty reads `icrnl ixon opost`
and the node produces garbage; forcing raw mode makes it read clean, single
reader, on the first start.

## Fix

Put the port in raw mode. The tty setup is factored into a small, dependency-free
`serial_port` library:

```cpp
bool configureSensorTty(int fd);  // cfmakeraw() + 115200 8N1 + non-blocking
```

`OpenAndConfigurePort()` now delegates to it. `cfmakeraw()` clears the
byte-mangling `c_iflag`/`c_oflag`/`c_lflag` bits regardless of the tty's prior
state, so the reader no longer depends on what a previous opener left behind.

## Testing

**Unit (hermetic, no hardware)** — `test_serial_port.cpp` drives a pseudo-terminal:
- `ForcesRawModeFromCooked` — start the PTY in cooked mode (`ICRNL`/`IXON`/`OPOST`
  set), run `configureSensorTty`, assert those flags are cleared. This is the
  regression guard for the bug above.
- `IdempotentOnRawTty`, `Sets8BitChars`.

**On hardware**
- 2/2 physical USB replugs: fresh tty came up cooked, fixed node read clean
  (`temporal_jit=0`, no rail hits) on the first start.
- 8/8 trials from a `stty`-forced cooked state: all clean.

## Note on the SDK-backed poller

The SDK poller reads this sensor cleanly for one reason only: libserialport sets
raw mode on open. The parsing, CRC handling, and init commands are otherwise
equivalent to `poll_data_node`. This one-line-of-behavior fix addresses the
flashing-red problem in the regular poller directly, so the SDK rewrite is not
required to solve it.

Out of scope: the sensor occasionally comes up *silent* (no stream at all) after
an unclean shutdown — that's absence of data, not garbage, affects any reader,
and is not addressed here.

### DIFF
--- a/robotiq_tsf/CMakeLists.txt
+++ b/robotiq_tsf/CMakeLists.txt
@@ -71,6 +71,17 @@ target_link_libraries(${PROJECT_NAME}_algorithms
     ${OpenCV_LIBRARIES}
 )
 
+# Serial-port helpers (raw-mode tty setup). Dependency-free so the unit tests
+# can link it without pulling in OpenCV/ROS.
+add_library(${PROJECT_NAME}_serial
+  src/serial_port.cpp
+)
+target_include_directories(${PROJECT_NAME}_serial
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>
+)
+
 add_executable(poll_data_node src/PollData.cpp)
 target_include_directories(poll_data_node
   PRIVATE
@@ -78,6 +89,7 @@ target_include_directories(poll_data_node
 )
 target_link_libraries(poll_data_node
   ${PROJECT_NAME}_algorithms
+  ${PROJECT_NAME}_serial
   ${cpp_typesupport_target}
   ${LIBUSB_LIBRARIES}
   ${OpenCV_LIBRARIES}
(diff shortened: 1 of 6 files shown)


mbegin-robotiq (approved):

ebarnett3:
closed, since we fix PollData.cpp here, and PollData.cpp is replaced in https://github.com/robotiq/ros/pull/15

PR #14, opened 2026-07-08
docs: PickNik ros2_robotiq_gripper migration guide

ebarnett3:
# docs: PickNik ros2_robotiq_gripper migration guide

Adds a **"Migrating from PickNik's ros2_robotiq_gripper"** section to the top-level `README.md` (under *Grippers*). Docs-only.

`grippers/` is the maintained continuation of PickNik's package; existing users need instructions to switch over. The new section covers:

- Drop-in framing — package/launch/controller/action names unchanged, so the PickNik clone must be removed first (duplicate-package `colcon` failure otherwise)
- Two migration paths: Docker (`./docker/run.sh gripper`) and existing Jazzy workspace (incl. a gripper-only build command)
- Breaking changes for `humble`-branch users (all upstream, pre-import): Humble→Jazzy, `.repos` rename, `GripperCommand`→`ParallelGripperCommand` with a before/after `send_goal` pair
- What you gain: `use_fake_hardware` arg, actionable comm errors (#12), unified Docker image

Facts checked against the tree (package.xml deps, launch args, `.repos`, `run.sh`, controller yaml). Validated end-to-end on real 2F-85 hardware: humble baseline container → this repo's Docker path and the workspace path (duplicate-package failure reproduced, both build variants pass); every command ran verbatim.

Validation surfaced one pre-existing driver bug (out of scope here, ticketed internally): `robotiq_driver` never updates its velocity state interface, so with the vendored config's `stall_timeout: 0.05` every `gripper_cmd` goal reports `stalled: true, reached_goal: false` while the gripper still physically completes the motion.

FYI @bcastets-robotiq @l-tschreiber-a11y @JonathanBeaudoinRq 

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,72 @@ ROS 2 `ros2_control` driver for Robotiq grippers (2F-85 / 2F-140, Hand-E), under
 
 `grippers/` is a vendored copy of PickNik Robotics' [`ros2_robotiq_gripper`](https://github.com/PickNikRobotics/ros2_robotiq_gripper) (BSD-3-Clause), imported via `git subtree` at upstream commit `3b6cf8f` with its history preserved, and maintained here in-tree. Upstream copyright and `<author>` tags are retained.
 
+### Migrating from PickNik's ros2_robotiq_gripper
+
+This repository is the maintained continuation of the PickNik package. Package names (`robotiq_driver`, `robotiq_controllers`, `robotiq_description`, `robotiq_hardware_tests`), launch files, controller names, and the `/robotiq_gripper_controller/gripper_cmd` action are all unchanged, so at the workspace level it is a drop-in replacement.
+
+**Docker (recommended):**
+
+```bash
+git clone --recurse-submodules https://github.com/robotiq/ros.git
+cd ros && ./docker/run.sh gripper
+```
+
+**Existing ROS 2 Jazzy workspace:**
+
+```bash
+cd ~/ws/src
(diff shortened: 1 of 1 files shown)


mbegin-robotiq (approved):

l-tschreiber-a11y:
why  switching from Humble→Jazzy, is related to  rename GripperCommand→ParallelGripperCommand ?

ebarnett3:
> why switching from Humble→Jazzy, is related to rename GripperCommand→ParallelGripperCommand ?

They're related but the distro switch isn't what forces it — and it's not a rename. The exact chain, from upstream PickNik [commit 0230e84 (PR #103)](https://github.com/PickNikRobotics/ros2_robotiq_gripper/pull/103): "Replace 'gripper_controllers' package that was removed in Kilted+ with 'parallel_gripper_controller' available in Jazzy+."

- control_msgs/action/ParallelGripperCommand is a new action type introduced alongside the new parallel_gripper_controller package in Jazzy (JointState-based goal, multi-joint capable, velocity/effort limits) — GripperCommand still exists in control_msgs.
- Strictly speaking, Jazzy alone doesn't force the change: both gripper_controllers (old, GripperCommand) and parallel_gripper_controller (new) are released in Jazzy.
- The forcing event is one distro later: gripper_controllers is removed in Kilted+. PickNik's main line therefore switched to the parallel controller, and Jazzy is the earliest distro where that replacement exists.
- Our repo continues PickNik main, so for someone migrating from the humble branch the two changes arrive together — same jump, but the action change comes from following the maintained line (and staying viable past Jazzy), not from Jazzy itself.

PR #15, opened 2026-07-13
feat(robotiq_tsf)!: replace poll_data_node with SDK-backed poll_data_…

ebarnett3:
## Summary

Replaces `poll_data_node` with `poll_data_sdk_node`, an SDK-backed driver for the TSF-85 tactile sensor: the vendor SDK (`extern/tactile_sensors` submodule) does transport + packet parsing, the node is just the ROS bridge — same topics, service, and Madgwick fusion as before.

**Supersedes #11** (`fix-tactile-raw-tty`), which patched the legacy poller in place; with the poller removed it becomes moot and will be deleted.

## Problem

`poll_data_node`'s in-tree parser intermittently emits rail-to-rail garbage `StaticData` on the TSF-85. The failure is **bistable per plug event** (a replug randomly yields a healthy or a garbage session; orientation/hub/enumeration ruled out), and the node also segfaults on Ctrl-C. Measured on hardware, same plug state:

| Poller | Taxel range at rest | Per-taxel temporal std |
|---|---|---|
| `poll_data_node` | 67 – 65533 | ~10,000 – 24,000 |
| `poll_data_sdk_node` | 9,872 – 21,534 (normal) | 4 – 9 |

The bytes arrive intact; the legacy parser misreads them. With the SDK poller reading the identical device state cleanly, there is no configuration where the old poller is the better choice.

## Changes

- `PollDataSdk.cpp` (ROS bridge) + `device_autodetect` (finds the sensor by USB descriptor / udev symlink) + `sdk_bridge` (SDK→msg mapping, extracted for testability) + `waitForStreaming()` (cycles the SDK stream if the sensor opens silent, then fails with a replug hint).
- `PollData.cpp` and its CMake target removed; launches' `poller:=` defaults to `poll_data_sdk_node`; README/`run.sh` updated.
- Docker: SDK COPY'd into the image (+ `COLCON_IGNORE`), `libserialport-dev`, sensor mapped at its **real** tty name (libserialport resolves via `/sys/class/tty`). CI: submodule checkout + `libserialport-dev` in the job container (no rosdep rule for the key).

## Testing

- Unit tests: `device_autodetect` (12 gtest, fixture `/dev` + sysfs trees) and `sdk_bridge` (3 gtest); all package suites pass via `colcon test`. `docker build` passes.
- Hardware: clean `StaticData` in plug states where the legacy parser emitted garbage; `EulerAngle` correct at rest and under physical rotation (yaw remains IMU-only/relative, as before); SIGINT while streaming → exit 0 (legacy segfaults there).

## Notes

- **Licensing**: the SDK sources are GPL-3.0 and are statically linked — now the only driver, so worth a conscious decision before shipping distributed binaries.
- **Known issue (host-only)**: outside the container the SDK constructor can block indefinitely (ModemManager suspected); `docker/run.sh` unaffected. To investigate separately.

FYI @JonathanBeaudoinRq @bcastets-robotiq


robotiq_tsf/include/robotiq_tsf/sdk_bridge.hpp
```diff
@@ -0,0 +1,18 @@
...
+#include "robotiq_tsf/msg/sensor.hpp"
+
+#include "finger_data.h"
+
+namespace robotiq_tsf
+{
+
+// Copy one SDK Fingers sample into the aggregate Sensor message: staticdata
+// taxels, dynamic, accelerometer, gyroscope and per-finger timestamps.
+// (EulerAngle/Quaternion are filled by the caller's AHRS fusion, not here.)
```
mbegin-robotiq, line 38: **`TactileSensor/Quaternion` is no longer published.** The legacy `poll_data_node` created `g_quat_pub` and published a `Quaternion` on every fused frame; `poll_data_sdk_node` declares no quaternion publisher and never computes one — only `EulerAngle` is published. This looks unintentional: this very comment still says "EulerAngle/**Quaternion** are filled by the caller's AHRS fusion", and the old code explicitly steered consumers to `Quaternion` for *continuous* (non-wrapping) orientation, since Euler wraps at ±180°. As a breaking PR, dropping a published topic should at least be called out in the description.

If the topic should stay (`MadgwickFilter::getQuaternion` still exists), please restore it. If the drop is intentional, update this comment and the README topic table.

Either way, **let's add a regression test that pins the node's published topic set** so this can't silently drift again — e.g. a small rostest/launch_test that brings the node up (SDK stubbed or `use_fake_hardware`-style) and asserts `TactileSensor/{StaticData,Dynamic,Accelerometer,Gyroscope,EulerAngle,Quaternion,Timestamp}` are all advertised. The current gtests cover `sdk_bridge`/`device_autodetect` well but nothing guards the ROS surface (topics/QoS), which is exactly what regressed here.

robotiq_tsf/src/poll_data_sdk.cpp
```diff
@@ -0,0 +1,392 @@
...
+{
+    frames_received_.fetch_add(1, std::memory_order_relaxed);
+    if (stopped_.load())
+        return;
+
+    robotiq_tsf::fillSensorMessages(fingers, sensors_data_);
+
+    // Measured dt for the filter integration (the SDK streams at ~700 Hz).
+    const auto now_upd = std::chrono::steady_clock::now();
+    float dt = 1.0f / 700.0f;  // fallback on the first sample
```
mbegin-robotiq: **`dt` is no longer clamped before it feeds the Madgwick integrator.** The legacy path bounded the integration step to `[dt_clamp_lo, dt_clamp_hi]` (~0.1 ms – 100 ms) so a delayed sample couldn't inject an outsized `dt`. Here `dt` comes straight from wall-clock callback arrival with no bound:

```cpp
float dt = 1.0f / 700.0f;
if (last_update_.time_since_epoch().count() != 0)
    dt = std::chrono::duration<float>(now_upd - last_update_).count();
```

A GC/scheduling stall, or the `stop()`→`start()` service cycle (and the stream-cycling in `waitForStreaming`), can produce a multi-hundred-ms `dt` that briefly corrupts the orientation estimate. It's self-correcting via the accel gate, but the clamp was cheap insurance — recommend restoring a `[lo, hi]` clamp (ideally keeping the `madgwick.dt_clamp_*` parameters).

**Please also add a unit test for the clamp** so this can't regress again. The clamp logic is easy to pull into a small free function (`float clampDt(float dt, float lo, float hi)`) and test directly — assert a normal `dt` passes through, an oversized/stalled `dt` is capped to `hi`, and a sub-threshold `dt` is raised to `lo`. That mirrors how `sdk_bridge`/`device_autodetect` were already factored for testability.

robotiq_tsf/src/poll_data_sdk.cpp
```diff
@@ -0,0 +1,392 @@
...
+void PollDataSdkNode::handleFingers(const Fingers &fingers)
+{
+    frames_received_.fetch_add(1, std::memory_order_relaxed);
+    if (stopped_.load())
+        return;
+
+    robotiq_tsf::fillSensorMessages(fingers, sensors_data_);
+
+    // Measured dt for the filter integration (the SDK streams at ~700 Hz).
+    const auto now_upd = std::chrono::steady_clock::now();
```
mbegin-robotiq: **Consider driving `dt` (and the message timestamps) from the sensor's hardware clock rather than the host `steady_clock`.** Two related points:

1. **`dt` for the fusion.** This uses one host-clock delta shared across both fingers. The legacy node integrated each finger against its own MCU timestamp (`finger.timestamp`, ms). The per-finger hardware timestamps are almost certainly the better source: they're immune to host scheduling jitter and reflect the actual sample interval the firmware produced, so a delayed callback doesn't distort the integration step. Since both fingers arrive in one packet the shared value is close in practice, but per-finger MCU deltas are the more correct input (and would also make the [#dt-clamp comment](../) largely moot for the common case).

2. **Published message timestamps.** `fillSensorMessages` already copies `finger.timestamp` into `out.timestamp.values[f]`, but if/when these messages carry a `header.stamp` (or downstream consumers correlate them), those stamps should be derived from the hardware timestamp, not `now()` — otherwise host latency/jitter smears the timeline that the firmware already timestamped precisely.

The main caveat is wrap/seeding of the MCU timestamp (the legacy code seeded `g_last_ts_ms[f]=0` after bias calc and skipped the first delta) — worth carrying that handling over. A small unit test on the per-finger `dt` derivation (normal delta, first-sample seed, backwards/zero delta) would lock it in.

robotiq_tsf/src/poll_data_sdk.cpp
```diff
@@ -0,0 +1,392 @@
...
+    float dt = 1.0f / 700.0f;  // fallback on the first sample
+    if (last_update_.time_since_epoch().count() != 0)
+        dt = std::chrono::duration<float>(now_upd - last_update_).count();
+    last_update_ = now_upd;
+
+    if (bias_iter_ > kBiasCalculationIterations)
```
mbegin-robotiq: **Two fusion-tuning behaviors from the legacy node are gone here — please confirm both are intentional.**

1. **Online gyro-bias trimming removed.** `poll_data_node` ran a still-detection loop after the initial bias calc (`still_gyro_eps_deg_s` / `still_accel_eps_g` gates + `bias_learn_rate` EMA) that slowly re-trimmed the gyro bias whenever the sensor was at rest. That whole path is absent now: bias is estimated once during the first `kBiasCalculationIterations` samples and then frozen. With yaw being IMU-only/relative this may be acceptable, but a frozen bias means faster yaw drift over long sessions and no recovery from thermal drift — worth a conscious call rather than a silent drop.

2. **`madgwick.*` parameters no longer declared.** The constructor only declares `device` and `publish_rate_hz`; the legacy `madgwick.beta`, `madgwick.accel_gate_*`, `madgwick.still_*`, `madgwick.bias_learn_rate`, `madgwick.dt_clamp_*` parameters are gone. Existing launch files / configs that still set these will now hit an "undeclared parameter" error at startup. If the tunables are intentionally dropped, note it in the PR (breaking-change surface); if not, re-declare at least `beta` and the accel gate.

If the online trim stays removed, a small unit test asserting the still-detection/bias-EMA behavior would be the natural regression guard once it's re-introduced — otherwise there's nothing pinning the fusion tuning surface.

robotiq_tsf/CMakeLists.txt
```diff
@@ -8,12 +8,31 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
+# parsing that the SDK-backed poller (poll_data_sdk_node) relies on. Built as a
+# small static lib from the single SDK translation unit.
+set(TSF_SDK_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../extern/tactile_sensors/sdk_cpp")
+if(NOT EXISTS "${TSF_SDK_DIR}/RobotiqTactileSensor.cpp")
+  message(FATAL_ERROR
+    "Tactile sensors SDK not found at ${TSF_SDK_DIR}. "
+    "Initialize the submodule ('git submodule update --init --recursive') or, "
+    "for the Docker build, ensure docker/Dockerfile COPYs it into the workspace.")
+endif()
+add_library(robotiq_tactile_sdk STATIC ${TSF_SDK_DIR}/RobotiqTactileSensor.cpp)
```
mbegin-robotiq, line 30: **Licensing sign-off needed before merge.** This statically links the `tactile_sensors` SDK, whose sources are GPL-3.0, into `poll_data_sdk_node` — now the *only* tactile driver, and it ships in the Docker image. Static linking of GPL code makes the resulting binary a derivative work subject to GPL-3.0 distribution terms (source availability, license propagation) for anything we hand out.

The author already flagged this in the PR Notes; raising it here as a resolvable thread so it isn't lost. This is a business/legal decision, not a code one — someone with authority over Robotiq's distribution/licensing should explicitly sign off before this merges if we distribute binaries. Options if that's a problem: dynamic-link/`dlopen` the SDK, isolate it behind a separate process/IPC, or confirm the SDK can be relicensed for our use. Resolve this thread once we have an explicit decision on record.

l-tschreiber-a11y (approved):

mbegin-robotiq (approved):
Approving — strong, well-structured refactor: parsing moved into the SDK, `sdk_bridge`/`device_autodetect` cleanly extracted and unit-tested, and the shutdown thread-handshake is correct. Solid work.

Approval is on the code. Two things I'd still like closed out before this actually merges (left as inline threads):
- **GPL-3.0 licensing** — static-linking the SDK into the only shipped driver needs an explicit distribution sign-off. Please resolve that thread with a decision on record.
- **Dropped `TactileSensor/Quaternion` topic** — confirm intentional (and update docs/comment) or restore it.

The `dt`-clamp / hardware-clock / bias-trim notes are non-blocking improvements. Nice PR overall.

robotiq_tsf/CMakeLists.txt
ebarnett3: Converted the PR to draft to prevent accidental merge before this issue is resolved.

robotiq_tsf/include/robotiq_tsf/sdk_bridge.hpp
ebarnett3: Done. Restored the publisher; `test_poll_data_sdk_node` now pins the advertised
topic set so it can't drift again.

robotiq_tsf/src/poll_data_sdk.cpp
ebarnett3: Done — both restored: the still-detection gyro-bias trim, and the `madgwick.*`
parameters (legacy defaults, applied to both filters). Trim helpers tested in
`test_fusion`.

robotiq_tsf/src/poll_data_sdk.cpp
ebarnett3: Done for the `dt`: per-finger MCU timestamp via `deriveDt`, with the seed/wrap
handling carried over; tested in `test_fusion`.

Point 2 (message timestamps): `fillSensorMessages` already copies
`finger.timestamp` into the `Timestamp` topic. No message has a `header.stamp`
field yet, so I've left deriving one from the hardware clock as a follow-up
rather than add an unused field now.

robotiq_tsf/src/poll_data_sdk.cpp
ebarnett3: Done — clamp and the `madgwick.dt_clamp_lo/hi` params restored, tested in
`test_fusion`. One deviation: applied via `std::clamp` inside `deriveDt` rather
than a standalone `clampDt` helper, since clamping isn't `dt`-specific.

robotiq_tsf/CMakeLists.txt
ebarnett3: This licensing issue was resolved in the tactile_sensors repo, in this PR: https://github.com/robotiq/tactile_sensors/pull/9

Robotiq has the rights to this code so we relicensed to BSD 3-clause.

PR #16, opened 2026-07-14
fix(robotiq_description): use exported interface names in controller …

ebarnett3:
Fixes upstream [PickNik issue #114](https://github.com/PickNikRobotics/ros2_robotiq_gripper/issues/114).

`robotiq_controllers.yaml` claims `<joint>/effort` and `<joint>/velocity`, but the hardware exports `set_gripper_max_effort` / `set_gripper_max_velocity`, so `robotiq_gripper_controller` fails to activate on source builds. Previously masked by a `sed` patch in `docker/Dockerfile` that rewrote the yaml inside the image only.

- Rename the two interfaces in `robotiq_controllers.yaml` to the exported names; drop the `sed` patch.
- Regression tests: `robotiq_driver` gtest pins the exported `set_gripper_max_*` names; new `robotiq_description` pytest asserts the yaml only claims exported interfaces (fails on the pre-fix yaml).

Verified on a real 2F-85: controllers activate, and per-goal speed reaches the gripper (full stroke ~0.6 s at `velocity: [0.15]` vs ~1.7 s at `[0.02]`), including in a docker image rebuilt without the patch. Fake-hardware activation remains broken for a different reason (mock hardware only creates URDF-declared interfaces) and will be addressed separately.

PTI @JonathanBeaudoinRq

### DIFF
--- a/.github/workflows/ci-ros-build-test.yml
+++ b/.github/workflows/ci-ros-build-test.yml
@@ -1,4 +1,5 @@
-# Build robotiq_tsf and run its unit tests (colcon test) inside a ROS container.
+# Build robotiq_tsf and the gripper packages and run their unit tests
+# (colcon test) inside a ROS container.
 # Running in ros:<distro>-ros-base gives a complete, consistent ROS environment.
 # The previous setup-ros-on-a-bare-runner approach intermittently produced an
 # incomplete ROS install (missing ament_package / rosidl_typesupport_c /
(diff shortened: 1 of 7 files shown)


ebarnett3:
reminder for this pull request

mbegin-robotiq (approved):

mbegin-robotiq:
merci !

l-tschreiber-a11y (approved):

PR #17, opened 2026-07-14
Sgrip 867/eigen linalg

ebarnett3:
## What

Migrates the AHRS/fusion code from hand-rolled quaternion and vector math to Eigen, with the numerics regression-pinned by tests recorded before the conversion.

- **test**: pre-migration hardening — Hamilton-convention component pinning, gimbal-lock clamp, combined roll+pitch init, golden-trajectory quaternion checkpoints, rpy→quat golden (Python viz).
- **chore**: delete dead `MahonyAHRS.{h,cpp}` (never built).
- **refactor**: `MadgwickFilter` state becomes `Eigen::Quaternionf`; gyro derivative, integration, normalization, and `initFromAccel` use Eigen. The Madgwick gradient-descent step stays as the reference scalar expansion. Free helpers `quatMul`/`quatConj`/`quatNormalize`/`quatFromAxis*` deleted (Eigen covers them); `quatToEulerRad/Deg` kept as thin helpers — Eigen's `eulerAngles(2,1,0)` constrains its first angle to `[0, π]`, which doesn't match the published aerospace ±180° ZYX convention.
- **refactor**: `fusion.hpp` and the poller vectorized with `Eigen::Vector3f`; the duplicated finger-0/finger-1 fusion blocks collapse into one `FINGER_COUNT` loop; new reusable `test/eigen_test_utils.hpp` (`expectEigenFloatEq`).

## Numeric parity

- Golden-trajectory checkpoints (recorded from the pre-Eigen implementation) pass unmodified at 1e-5 tolerance; accel normalization deliberately keeps reciprocal-multiply to match the reference rounding (plain division drifted the trajectory by ~1e-4).
- `initFromAccel` keeps the original norm-vs-eps degenerate-input guard.

## Testing

- 61/61 unit tests green on a clean from-scratch Release build.
- Hardware check (TSF-85): both orientation topics publish after bias calibration; quaternions unit-norm to 1e-6; Euler/quaternion self-consistent on both fingers; stationary drift over 5 s ~1.3°/0.6° (normal gyro-only yaw drift).
FYI @jkski-robotiq

### DIFF
--- a/robotiq_tsf/src/MadgwickAHRS.cpp
+++ b/robotiq_tsf/src/MadgwickAHRS.cpp
@@ -1,3 +1,31 @@
+// Copyright (c) 2026 Robotiq
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//    * Redistributions of source code must retain the above copyright
+//      notice, this list of conditions and the following disclaimer.
+//
+//    * Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//    * Neither the name of the copyright holder nor the names of its
+//      contributors may be used to endorse or promote products derived from
+//      this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
 // MadgwickAHRS.cpp
 //
 // Class-based Madgwick IMU AHRS filter (gyro+accel only). Derived from the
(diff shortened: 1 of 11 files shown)


mbegin-robotiq:
super merci !

robotiq_tsf/src/MadgwickAHRS.cpp
```diff
@@ -49,180 +48,113 @@ void MadgwickFilter::initFromAccel(float ax, float ay, float az) {
...
+            // implementation's rounding exactly.
+            const Eigen::Vector3f an = accel*(1.0f/accelNorm);
+
+            // Madgwick gradient-descent step, kept as the reference scalar
+            // expansion (it is the algorithm itself, not generic algebra).
+            const float q0 = q_.w();
+            const float q1 = q_.x();
+            const float q2 = q_.y();
+            const float q3 = q_.z();
+            const float _2q0 = 2.0f*q0;
```
mbegin-robotiq: **Formatting — resolve before merge.** This tight operator spacing (`2.0f*q0`, and similarly `accel*(1.0f/accelNorm)`, `beta_*s.coeffs()`, `gyro*deg_to_rad`) won't survive clang-format.

`robotiq_tsf` had no format enforcement until now, which is why it slipped through. #18 adds `robotiq_tsf/.clang-format` (monorepo/`grippers` Google style), a clang-format-19 pre-commit hook, and a `robotiq_tsf` CI job, plus a one-time reflow.

Once #18 lands: rebase this branch (and its base `tactile-sdk-poller`) onto `main`, then run `clang-format-19 -i` over the touched files (or `pre-commit run --config robotiq_tsf/.pre-commit-config.yaml -a`) and commit. Expect conflicts in `MadgwickAHRS.{h,cpp}` — keep this PR's content, then re-run clang-format. The new `Format (robotiq_tsf)` check will gate the merge. Resolve this thread once formatted.

robotiq_tsf/test/test_madgwick_ahrs.cpp
```diff
@@ -108,32 +108,63 @@ TEST(QuatHelpers, MulByConjugateGivesRelativeRotation) {
...
+    EXPECT_NEAR(q_rel.x(), 0.0f, kQuatTol);
+    EXPECT_NEAR(q_rel.y(), 0.0f, kQuatTol);
+    EXPECT_NEAR(q_rel.z(), 0.0f, kQuatTol);
 }
 
-TEST(QuatHelpers, NormalizeScalesToUnitAndHandlesZero) {
-    float q[4] = {2.0f, 0.0f, 0.0f, 0.0f};
-    quatNormalize(q);
-    EXPECT_NEAR(q[0], 1.0f, kQuatTol);
+TEST(QuatHelpers, HamiltonConventionGoldenComponents) {
```
mbegin-robotiq: **Nit (convention, non-blocking):** the tests here use `TEST(Suite, PascalCase)` rather than the `given_when_then` naming the Robotiq C++ testing conventions set as the default (§2, "Test Naming").

Pre-existing in these files, not introduced by this PR — so no need to churn it here. Flagging so it's a conscious choice: the conventions do allow a pragmatic exception ("for simple, single-function or data-driven tests, a direct name like `MethodName_WithSpecificInput` is acceptable"), which arguably covers these pure-math cases. Fine to resolve as won't-fix, or fold a rename into a later cleanup.

mbegin-robotiq (approved):
Approving — the Eigen migration is correct and well-protected: I traced the Hamilton-product gyro derivative, `initFromAccel`, and the gradient step against the originals, and the golden-trajectory + convention-pinning tests lock in numeric parity. The finger-loop consolidation is a clean dedup.

Non-blocking follow-ups left as resolvable threads (formatting, DRY test helper, test naming). One pre-merge must-do: apply clang-format once #18 lands (rebase + `clang-format-19 -i`) so the new `Format (robotiq_tsf)` check passes. Not merging — leaving that to you.

robotiq_tsf/src/MadgwickAHRS.cpp
ebarnett3: Done

robotiq_tsf/test/test_madgwick_ahrs.cpp
ebarnett3: Resolving as won't-fix. These are pure-math / data-driven cases, which the
conventions explicitly allow to name directly (§2 exception:
"`MethodName_WithSpecificInput` is acceptable for simple, single-function or
data-driven tests"). The `TEST(Suite, PascalCase)` naming here is a conscious
choice, not an oversight.

l-tschreiber-a11y (approved):

PR #1, opened 2026-03-05
Updates 106

jkski-robotiq:
updates to ros package and docker file. uopdates to readme. should work with 1.0.6 and uint16 timestamp

PR #20, opened 2026-07-27
feat: grippers sdk integration

ebarnett3:
Corresponding PR in the grippers repo: https://github.com/robotiq/grippers/pull/7

PR 1 of 2

`robotiq_driver` stops running its own Modbus stack and drives
`Robotiq::Gripper` from a new `extern/grippers` submodule. Two commits: 

1. The SDK arrives unused (submodule, build wiring, adapters)
2. The hardware interface switches over

The SDK is compiled into `librobotiq_driver.so` rather than built as a second
workspace package. Its headers
ship with the package, so a consumer may call SDK entry points the wrapper
itself never references. Docker and CI build the submodule exactly as they do
the tactile SDK. The gripper  package's `format`/`tidy` convenience targets are
dropped: they collided with the SDK's, and pre-commit is this repository's
formatting authority anyway.

Three small adapters land with the wiring, each testable without a gripper, a
serial port or a controller manager, sharing one `robotiq_driver_tests`
binary: 

- `rclcpp_logger` (SDK diagnostics → `/rosout` at matching severity, format-string-safe), 
- `hardware_parameters` (`HardwareInfo` →`Robotiq::ConnectionConfig`, with the
   defaults pinned here so the SDK changing its own cannot break an existing description)
- `gripper_scaling` (the register/joint arithmetic — the part that silently mis-commands a
   gripper when it is wrong).

The switch itself collapses the wrapper to unit conversion and lifecycle. The
SDK keeps a whole-snapshot process image behind a mutex, refreshed by its own
exchange thread, so `read()` and `write()` are pure memory operations that can
no longer stall the controller manager; the hand-rolled framing layer, CRC
table, serial abstraction, factories and `background_task()` thread all become
dead code (deleted in PR 2). `FakeDriver` becomes the SDK's
`makeFakeGripper()`, and link health is logged through `connectionState()`
instead of being swallowed.

## Follow-ups

Deliberately held back so the migration stays a migration; the first four pair
with the *Deliberately preserved bugs* table in the break analysis.

1. Conservative activation/deactivation behind a parameter — today either
   transition **drops whatever the gripper is holding**; the highest-value item.
2. Round the register mapping instead of truncating (one count low, 1 in 227).
3. Report recovery failure on `reactivate_gripper_response`.
4. Populate the `velocity` state interface (has always read 0), or expose gCU.
5. Decide what a faulted link does to the component once SDK reconnection lands.
6. Plumb `connection_frequency`/`activation_timeout` through the xacro macros.
7. Move the SI → register conversion into the SDK (SGRIP-872).
8. Non-blocking forms of the SDK's blocking procedures (SGRIP-873).
9. Move the ROS-free tests into the SDK (SGRIP-874).
10. Re-pin the submodule to a release tag once the SDK train lands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,8 @@ rm -rf ros2_robotiq_gripper        # remove the PickNik clone to prevent duplica
 git clone --recurse-submodules https://github.com/robotiq/ros.git
 vcs import < ros/grippers/ros2_robotiq_gripper.rolling.repos   # same external `serial` dep as before
 cd ~/ws
-rosdep install --from-paths src --ignore-src -y
+sudo apt install libserialport-dev # the gripper SDK's serial backend; no rosdep key exists for it
+rosdep install --from-paths src --ignore-src -y --skip-keys libserialport
 rm -rf build install               # clear artifacts built from the PickNik sources
 colcon build
 ```
@@ -236,12 +237,11 @@ Two SDKs live under `extern/` as submodules and are built into the image:
 | Submodule | Repository | Used by |
 |---|---|---|
 | `extern/tactile_sensors` | [Robotiq/tactile_sensors](https://github.com/Robotiq/tactile_sensors) | `robotiq_tsf`'s `poll_data_sdk_node` |
-| `extern/grippers` | [Robotiq/grippers](https://github.com/Robotiq/grippers) | the gripper C++ SDK |
+| `extern/grippers` | [Robotiq/grippers](https://github.com/Robotiq/grippers) | `robotiq_driver` |
 
 `extern/COLCON_IGNORE` keeps colcon from building either SDK's standalone
 CMake project as a workspace package; the packages that need them compile them
 in-tree.
-
 
 | Script | Description |
 |---|---|
--- a/README.md
+++ b/README.md
@@ -231,6 +231,18 @@ If you already cloned without `--recurse-submodules`:
 git submodule update --init
 ```
 
+Two SDKs live under `extern/` as submodules and are built into the image:
+
+| Submodule | Repository | Used by |
+|---|---|---|
+| `extern/tactile_sensors` | [Robotiq/tactile_sensors](https://github.com/Robotiq/tactile_sensors) | `robotiq_tsf`'s `poll_data_sdk_node` |
+| `extern/grippers` | [Robotiq/grippers](https://github.com/Robotiq/grippers) | the gripper C++ SDK |
+
+`extern/COLCON_IGNORE` keeps colcon from building either SDK's standalone
+CMake project as a workspace package; the packages that need them compile them
+in-tree.
+
+
 | Script | Description |
 |---|---|
 | `run.sh` | Builds a single ROS 2 Jazzy image with the **whole workspace** (sensor + grippers) and launches a shell with that product's devices mapped: `./run.sh [gripper\|sensor\|both]`. |
--- a/grippers/robotiq_driver/CMakeLists.txt
+++ b/grippers/robotiq_driver/CMakeLists.txt
@@ -26,6 +26,34 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
   serial
 )
 
+###############################################################################
+# GRIPPER SDK
+
+# The Modbus transport, the typed register blocks and the Robotiq::Gripper
+# runtime API live in the extern/grippers submodule. Nothing links against it
+# yet — the hardware interface still runs the in-tree driver — but the build
+# wiring and the adapters below land first so the switch is a diff about
+# behaviour and nothing else.
+#
+# EXCLUDE_FROM_ALL skips the SDK's own install/export rules: it is compiled
+# into robotiq_driver, not shipped as a second installed package.
+set(GRIPPERS_SDK_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../extern/grippers/sdk_cpp")
+if(NOT EXISTS "${GRIPPERS_SDK_DIR}/CMakeLists.txt")
+  message(FATAL_ERROR
+    "The gripper SDK is missing from ${GRIPPERS_SDK_DIR}. "
+    "Run `git submodule update --init extern/grippers` at the repository root.")
+endif()
+# The SDK declares its library without STATIC or SHARED, so it follows
+# BUILD_SHARED_LIBS. Nothing installs it, so a shared one would leave the
+# plugin needing a library that does not ship. Pinned for this subtree only;
+# the enclosing project keeps its own setting.
+set(BUILD_SHARED_LIBS OFF)
+add_subdirectory(${GRIPPERS_SDK_DIR} ${CMAKE_CURRENT_BINARY_DIR}/grippers_sdk EXCLUDE_FROM_ALL)
+unset(BUILD_SHARED_LIBS)
+# A static archive going into a shared plugin has to be position-independent.
+set_target_properties(grippers_sdk PROPERTIES POSITION_INDEPENDENT_CODE ON)
+
+###############################################################################
 # Robotiq driver library.
 
 add_library(
@@ -41,7 +69,10 @@ add_library(
   include/robotiq_driver/driver.hpp
   include/robotiq_driver/driver_exception.hpp
   include/robotiq_driver/driver_factory.hpp
+  include/robotiq_driver/gripper_scaling.hpp
   include/robotiq_driver/hardware_interface.hpp
+  include/robotiq_driver/hardware_parameters.hpp
+  include/robotiq_driver/rclcpp_logger.hpp
   include/robotiq_driver/serial.hpp
   include/robotiq_driver/serial_factory.hpp
   include/robotiq_driver/visibility_control.hpp
@@ -53,12 +84,18 @@ add_library(
   src/fake/fake_driver.cpp
   src/default_serial.cpp
   src/default_serial_factory.cpp
+  src/hardware_parameters.cpp
+  src/rclcpp_logger.cpp
 )
 target_link_libraries(robotiq_driver atomic)
 target_include_directories(
   robotiq_driver
   PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  # The SDK's headers are part of this package's public API: the adapters here
+  # take Robotiq types, so downstream compilations have to see them too. They
+  # are installed next to ours below, hence the single install interface.
+  $<BUILD_INTERFACE:${GRIPPERS_SDK_DIR}/include>
   $<INSTALL_INTERFACE:include>
 )
 target_link_libraries(
@@ -68,6 +105,13 @@ target_link_libraries(
   rclcpp::rclcpp
   rclcpp_lifecycle::rclcpp_lifecycle
   serial::serial
+  # $<BUILD_INTERFACE:...> keeps the statically-linked SDK out of the installed
+  # export set, where there would be no target for it to resolve to.
+  # WHOLE_ARCHIVE (CMake 3.24+, satisfied from Jazzy's 24.04 on) because the
+  # SDK's headers ship with this package: a consumer may call any SDK entry
+  # point, so every SDK object has to land in the library — not just the
+  # ones the wrapper itself references.
+  $<BUILD_INTERFACE:$<LINK_LIBRARY:WHOLE_ARCHIVE,Robotiq::grippers>>
 )
 
 ###############################################################################
@@ -111,6 +155,14 @@ install(
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}          # include
 )
 
+# The SDK headers ship alongside ours: they are reachable from this package's
+# public headers, so a consumer including them needs these too.
+install(
+    DIRECTORY ${GRIPPERS_SDK_DIR}/include/
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}          # include
+    FILES_MATCHING PATTERN "*.hpp"                   # not the SDK's own CMakeLists
+)
+
 # Install our library.
 install(
   TARGETS robotiq_driver
@@ -130,15 +182,5 @@ include(CTest)
 if(BUILD_TESTING)
   add_subdirectory(tests)
 endif()
-
-###############################################################################
-# LINTERS
-
-add_custom_target(format
-    COMMAND clang-format -i `git ls-files *.hpp *.cpp`
-    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
-add_custom_target(tidy
-    COMMAND clang-tidy -p ${CMAKE_BINARY_DIR} `git ls-files *.cpp`
-    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
 
 ament_package()
--- a/grippers/robotiq_driver/CMakeLists.txt
+++ b/grippers/robotiq_driver/CMakeLists.txt
@@ -30,10 +30,8 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
 # GRIPPER SDK
 
 # The Modbus transport, the typed register blocks and the Robotiq::Gripper
-# runtime API live in the extern/grippers submodule. Nothing links against it
-# yet — the hardware interface still runs the in-tree driver — but the build
-# wiring and the adapters below land first so the switch is a diff about
-# behaviour and nothing else.
+# runtime API live in the extern/grippers submodule; the hardware interface
+# below drives the gripper through it.
 #
 # EXCLUDE_FROM_ALL skips the SDK's own install/export rules: it is compiled
 # into robotiq_driver, not shipped as a second installed package.
@@ -45,12 +43,13 @@ if(NOT EXISTS "${GRIPPERS_SDK_DIR}/CMakeLists.txt")
 endif()
 # The SDK declares its library without STATIC or SHARED, so it follows
 # BUILD_SHARED_LIBS. Nothing installs it, so a shared one would leave the
-# plugin needing a library that does not ship. Pinned for this subtree only;
-# the enclosing project keeps its own setting.
+# plugin needing a library that does not ship.
+set(_robotiq_saved_build_shared_libs "${BUILD_SHARED_LIBS}")
 set(BUILD_SHARED_LIBS OFF)
 add_subdirectory(${GRIPPERS_SDK_DIR} ${CMAKE_CURRENT_BINARY_DIR}/grippers_sdk EXCLUDE_FROM_ALL)
-unset(BUILD_SHARED_LIBS)
+set(BUILD_SHARED_LIBS "${_robotiq_saved_build_shared_libs}")
 # A static archive going into a shared plugin has to be position-independent.
+# CMake cannot set properties through the Robotiq::grippers alias.
 set_target_properties(grippers_sdk PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
 ###############################################################################
@@ -111,6 +110,11 @@ target_link_libraries(
   # SDK's headers ship with this package: a consumer may call any SDK entry
   # point, so every SDK object has to land in the library — not just the
   # ones the wrapper itself references.
+  #
+  # The consequence to know about: the whole SDK symbol set is exported from a
+  # dlopen'd plugin. A second package in the same process linking the SDK the
+  # same way gets its own copy with independent static state, and Robotiq::Gripper
+  # owns a serial link and a thread. One SDK consumer per process.
   $<BUILD_INTERFACE:$<LINK_LIBRARY:WHOLE_ARCHIVE,Robotiq::grippers>>
 )
 
@@ -157,6 +161,10 @@ install(
 
 # The SDK headers ship alongside ours: they are reachable from this package's
 # public headers, so a consumer including them needs these too.
+#
+# The pattern assumes the SDK is .hpp-only, which it is today. Anything else it
+# starts shipping — a C shim, a generated config header — would be dropped here
+# and fail at a downstream consumer's include, not in this build.
 install(
     DIRECTORY ${GRIPPERS_SDK_DIR}/include/
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}          # include
--- a/grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
+++ b/grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
@@ -38,6 +38,7 @@
 #include <algorithm>
 #include <cmath>
 #include <cstdint>
+#include <limits>
 #include <optional>
 
 namespace robotiq_driver {
@@ -77,18 +78,18 @@ inline constexpr uint8_t kGripperRange = kGripperMaxPos - kGripperMinPos;
    return static_cast<uint8_t>(std::clamp(counts, 0.0, 255.0));
 }
 
-//! An SI speed or force command -> its 0x00..0xFF register (rSP / rFR),
-//! as the fraction of \p maximum it represents. Sign is ignored: the
-//! registers set a magnitude, direction comes from the position request.
-//! Truncates, as above. Nothing when the fraction cannot be computed —
-//! either side non-finite, or a \p maximum that is not positive.
+//! An SI speed or force command -> its 0x00..0xFF register (rSP / rFR), as the
+//! fraction of \p maximum it represents. Truncates, as above. Nothing unless
+//! both sides are in range: \p value from zero up, \p maximum above it. The
+//! range excludes NaN and infinity, which no comparison admits.
 [[nodiscard]] inline std::optional<uint8_t> registerFromFractionOf(double value, double maximum)
 {
-   if(!std::isfinite(value) || !std::isfinite(maximum) || maximum <= 0.0)
+   constexpr double kLargest = std::numeric_limits<double>::max();
+   const bool in_range = value >= 0.0 && value <= kLargest && maximum > 0.0 && maximum <= kLargest;
+   if(!in_range)
    {
       return std::nullopt;
    }
-   const double fraction = std::clamp(std::fabs(value) / maximum, 0.0, 1.0);
-   return static_cast<uint8_t>(fraction * 0xFF);
+   return static_cast<uint8_t>(std::min(value / maximum, 1.0) * 0xFF);
 }
 } // namespace robotiq_driver
--- a/grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
+++ b/grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
@@ -38,6 +38,7 @@
 #include <algorithm>
 #include <cmath>
 #include <cstdint>
+#include <optional>
 
 namespace robotiq_driver {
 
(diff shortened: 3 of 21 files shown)


extern/grippers
mbegin-robotiq, line 1: a submodule works for now. but we should think about other strategies as well. I find the "Vendor/Wrapper Package" to be interesting (least effort while also not requiring users to use recursive clones).

[ROS C++ SDK Integration Strategies.pdf](https://github.com/user-attachments/files/30468701/ROS.C%2B%2B.SDK.Integration.Strategies.pdf)

grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
```diff
@@ -0,0 +1,82 @@
...
+//! Joint position -> register count (rPR), in the same unit as
+//! \p closed_position. Counts outside the byte are clamped; the caller is
+//! responsible for rejecting a NaN command.
+//!
+//! Truncates rather than rounds, so a position read off the gripper and
+//! commanded straight back can land one count below where it came from.
+[[nodiscard]] inline uint8_t registerFromJointPosition(double joint_position, double closed_position)
+{
+   const double counts = (joint_position / closed_position) * kGripperRange + kGripperMinPos;
+   return static_cast<uint8_t>(std::clamp(counts, 0.0, 255.0));
```
mbegin-robotiq: **Undefined behaviour when `closed_position` is 0.**

`parseParameters` accepts any parsable `gripper_closed_position`, including `0`. Then with `joint_position == 0.0` this computes `0.0/0.0 = NaN`. `std::clamp(NaN, 0.0, 255.0)` returns **NaN** — neither `comp(v, lo)` nor `comp(hi, v)` is true for a NaN — and `static_cast<uint8_t>(NaN)` is undefined. In practice that's an arbitrary `rPR` written straight to a real gripper.

The doc comment says "the caller is responsible for rejecting a NaN command", and `write()` does guard the command — but it can't guard against the *parameter* being zero, because that's baked into the divisor here.

Suggest rejecting a non-finite or `<= 0` `gripper_closed_position` in `parseParameters` (it already throws for an absent one, so this is the same class of fatal), and/or making the cast NaN-safe:

```cpp
const double counts = (joint_position / closed_position) * kGripperRange + kGripperMinPos;
return std::isfinite(counts) ? static_cast<uint8_t>(std::clamp(counts, 0.0, 255.0)) : kGripperMinPos;
```

grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+//! An SI speed or force command -> its 0x00..0xFF register (rSP / rFR),
+//! as the fraction of \p maximum it represents. Sign is ignored: the
+//! registers set a magnitude, direction comes from the position request.
+//! Truncates, as above.
+[[nodiscard]] inline uint8_t registerFromFractionOf(double value, double maximum)
+{
+   const double fraction = std::clamp(std::fabs(value) / maximum, 0.0, 1.0);
```
mbegin-robotiq: Same NaN-cast hazard as `registerFromJointPosition` above, via two routes:

- `maximum == 0` — a URDF setting `gripper_max_speed`/`gripper_max_force` to `0` parses fine today and gives `0.0/0.0 = NaN`.
- `value` itself NaN — see the note on `write()`, which only NaN-guards the position command.

`std::clamp` propagates the NaN rather than pinning it, so the `static_cast<uint8_t>` is undefined in both cases. Worth validating both maxima in `parseParameters` (fall back to the `kMaxSpeedDefault`/`kMaxForceDefault` constants rather than throwing — unlike the closed position, these *do* have a defensible default) and guarding the cast.

grippers/robotiq_driver/src/hardware_parameters.cpp
```diff
@@ -0,0 +1,137 @@
...
+   parameters.connection.modbusSlaveAddress =
+      parameterOr<uint8_t>(info, logger, kSlaveAddressParam, kSlaveAddressDefault, [](const std::string& text) {
+         return static_cast<uint8_t>(std::stoul(text, nullptr, 16));
+      });
+   parameters.connection.connectionFrequency =
+      parameterOr<double>(info, logger, kConnectionFrequencyParam, kConnectionFrequencyDefault, asDouble);
+
+   // No default: the joint mapping is meaningless without the gripper's
+   // closed angle, and guessing one would silently mis-scale every command.
+   parameters.closed_position = std::stod(info.hardware_parameters.at(kClosedPositionParam));
```
mbegin-robotiq, line 131: The rationale for having no default is right, but the validation stops one step short: `std::stod` happily returns `0.0`, `-0.5` or a denormal, and any of those makes the joint mapping either UB (see the `gripper_scaling.hpp` thread) or silently inverted.

Since this parameter is already fatal-on-absent, making it fatal-on-nonsensical is consistent:

```cpp
parameters.closed_position = std::stod(info.hardware_parameters.at(kClosedPositionParam));
if(!std::isnormal(parameters.closed_position) || parameters.closed_position <= 0.0)
{
   throw std::invalid_argument("gripper_closed_position must be a positive, finite joint value");
}
```

`on_init` already catches `std::exception` and returns `ERROR`, so this needs no call-site change.

grippers/robotiq_driver/src/hardware_parameters.cpp
```diff
@@ -0,0 +1,137 @@
...
+      logger,
+      kActivationTimeoutParam,
+      parameters.activation_timeout,
+      [](const std::string& text) {
+         return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::duration<double>(std::stod(text)));
+      });
+
+   // Anything but the literal "false" enables the dummy, "0" included.
+   parameters.use_dummy = info.hardware_parameters.count(kUseDummyParam) != 0
+                       && info.hardware_parameters.at(kUseDummyParam) != kUseDummyDefault;
```
mbegin-robotiq: **Anything but the literal `"false"` enables the fake gripper** — `"False"`, `"FALSE"`, `"0"`, `"no"` and `""` all included. `test_hardware_parameters.cpp` pins that behaviour explicitly, so it's deliberate, but the failure mode is bad: a capitalisation slip in a description runs a *simulated* gripper against production code, and the only signal is one `RCLCPP_INFO` line in `create_gripper()`.

The asymmetry is what makes it dangerous — every other parameter here fails safe toward the real default, this one fails toward the fake.

Suggest accepting the usual falsey spellings:

```cpp
const auto entry = info.hardware_parameters.find(kUseDummyParam);
parameters.use_dummy = entry != info.hardware_parameters.end() && !isFalsey(entry->second);
// isFalsey: "", "0", "false"/"False"/"FALSE" (case-insensitive compare)
```

And consider promoting the "connected to a dummy driver" line in `create_gripper()` from `INFO` to `WARN` — it's the kind of thing you want to notice in a log you're skimming.

grippers/robotiq_driver/src/hardware_parameters.cpp
```diff
@@ -0,0 +1,137 @@
...
+      });
+   // The URDF spells the timeout in seconds; SerialConfig wants milliseconds.
+   parameters.connection.serial.timeout =
+      parameterOr<std::chrono::milliseconds>(info, logger, kTimeoutParam, kTimeoutDefault, [](const std::string& text) {
+         return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::duration<double>(std::stod(text)));
+      });
+   // Base 16: the address is written as the manual prints it ("0x9").
+   parameters.connection.modbusSlaveAddress =
+      parameterOr<uint8_t>(info, logger, kSlaveAddressParam, kSlaveAddressDefault, [](const std::string& text) {
+         return static_cast<uint8_t>(std::stoul(text, nullptr, 16));
```
mbegin-robotiq: Base 16 unconditionally means `slave_address` of `"10"` is parsed as **16**, not 10. The comment justifies it from the manual's `0x9` notation, and the test pins both spellings — but base `0` gets you the manual's hex *and* reads a bare decimal as decimal, with no ambiguity either way:

```cpp
return static_cast<uint8_t>(std::stoul(text, nullptr, 0));
```

Note `gripper_interface_test.cpp:127` in this same PR switched its `--slave-address` handler to exactly that (`std::stoi(value, nullptr, 0)`), so the two paths currently disagree on what `10` means. Worth aligning.

Minor, same line: `std::stoul` silently truncates on the cast, so `"0x1FF"` becomes `0xFF` and a negative wraps. A range check before narrowing would make a typo loud instead of targeting the wrong device.

grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -42,33 +47,61 @@
...
-   }
+   return recovery.valid() && recovery.wait_for(std::chrono::seconds{0}) == std::future_status::ready;
 }
 
-// This constructor is use for testing only.
-RobotiqGripperHardwareInterface::RobotiqGripperHardwareInterface(std::unique_ptr<DriverFactory> driver_factory)
-   : driver_factory_{std::move(driver_factory)}
+// The throttled diagnostics below are wall-clock paced and independent of any
+// node, so they use their own clock rather than a lifecycle node's.
+rclcpp::Clock& diagnostic_clock()
```
mbegin-robotiq: This clock is a file-scope singleton, so `RCLCPP_WARN_THROTTLE` state is shared by **every** `RobotiqGripperHardwareInterface` in the process. On a dual-gripper cell, gripper A's link warning suppresses gripper B's for the full 5 s window — and since the messages don't name which gripper they're about, you can't tell from the log that it happened.

Making it a member (`rclcpp::Clock diagnostic_clock_{RCL_STEADY_TIME};`) fixes both halves. Including the port or the component name in the two throttled messages would help too.

grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -266,42 +312,104 @@ hardware_interface::CallbackReturn RobotiqGripperHardwareInterface::on_deactivat
...
+      if(recovery_.valid())
+      {
+         RCLCPP_WARN(kLogger, "A gripper recovery is already running; ignoring the reactivation request.");
+      }
+      else
+      {
+         RCLCPP_INFO(kLogger,
+                     "Recovering the Robotiq gripper: the reset releases any grip and sweeps the fingers through "
+                     "their full range.");
+         recovery_ = std::async(std::launch::async, [this] {
```
mbegin-robotiq, line 401: Spawning a thread from inside `read()` puts an allocation and a `clone()` on the control loop. It's request-gated so it won't happen every cycle, and the cost is tens of microseconds — but the header's headline claim is that "`read()` and `write()` are pure memory operations", and this is the one place that isn't.

Not a blocker. If you want to close the gap later, a persistent worker thread fed a `std::packaged_task` keeps the same future-shaped API with no allocation at request time.

grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -310,48 +418,33 @@ hardware_interface::return_type RobotiqGripperHardwareInterface::read(const rclc
...
-      {
-         RCLCPP_ERROR(kLogger, "Error: %s", e.what());
-      }
+      return hardware_interface::return_type::OK;
+   }
 
-      std::this_thread::sleep_for(kGripperCommsLoopPeriod);
+   // Before a controller has written a setpoint there is nothing to command,
+   // and a NaN would convert to a garbage register value.
+   if(std::isnan(gripper_position_command_))
```
mbegin-robotiq, line 444: This guards the position command, but `gripper_speed_` and `gripper_force_` two lines below go into `registerFromFractionOf` unchecked. They're seeded in `export_command_interfaces()` so they're never NaN at startup — but they're *command interfaces*, and nothing stops a controller writing NaN to `set_gripper_max_velocity`. That lands on the same undefined `static_cast<uint8_t>` (see the `gripper_scaling.hpp` threads).

Guarding inside `registerFromFractionOf` covers this and the zero-maximum case in one place, which is probably better than widening the condition here.

grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -242,18 +290,16 @@ hardware_interface::CallbackReturn RobotiqGripperHardwareInterface::on_activate(
       gripper_position_command_ = 0;
    }
 
-   // Activate the gripper.
-   try
+   // recoverFromFault(), not activate(): the manual's reset handshake, which
+   // clears a latched fault and runs the calibration sweep. Activating
+   // therefore releases any grip and sweeps the fingers.
+   const Robotiq::ActivationResult result = Robotiq::recoverFromFault(*gripper_, parameters_.activation_timeout);
```
mbegin-robotiq, line 303: Flagging this for visibility rather than as a change request, since it's follow-up #1 in the description and honestly documented right here in the comment.

This makes **every activation drop whatever the gripper is holding** and sweep the fingers — and `on_activate` is reached on any controller-manager restart, not just a deliberate reset. Combined with the up-to-15 s block, that's the highest-consequence behaviour in the PR.

It's pre-existing behaviour, so it isn't a regression and shouldn't hold up the migration. But I'd argue the conservative-activation parameter should land before this is deployed to anything holding a real part, rather than merely before the next release. Worth agreeing on that ordering explicitly now.

grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -142,43 +179,51 @@ hardware_interface::CallbackReturn RobotiqGripperHardwareInterface::on_init(
...
+   const rclcpp_lifecycle::State& /*previous_state*/)
 {
-   RCLCPP_DEBUG(kLogger, "on_configure");
-   try
-   {
-      if(hardware_interface::SystemInterface::on_configure(previous_state) != CallbackReturn::SUCCESS)
-      {
-         return CallbackReturn::ERROR;
-      }
+   RCLCPP_DEBUG(kLogger, "on_cleanup");
```
mbegin-robotiq, line 210: `on_cleanup` is thorough, but there's no `on_shutdown` or `on_error` override — so an error transition leaves `gripper_` and its exchange thread alive until the component is destroyed, and a still-running `recovery_` is never awaited on that path.

The destructor saves you (the `recovery_`-after-`gripper_` member ordering is correct, and `std::async`'s future blocks on destruction), so this is a tidiness point rather than a leak. An `on_shutdown` delegating to the same body would be cheap insurance.

Separately: the member-ordering invariant is documented in the header, which is good — but a reorder would break it silently at runtime with no compile error. Worth restating in the class-level docs, since that's where someone adding a member will be looking.

grippers/robotiq_driver/CMakeLists.txt
```diff
@@ -26,6 +26,34 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
   serial
 )
 
+###############################################################################
+# GRIPPER SDK
+
+# The Modbus transport, the typed register blocks and the Robotiq::Gripper
+# runtime API live in the extern/grippers submodule. Nothing links against it
+# yet — the hardware interface still runs the in-tree driver — but the build
+# wiring and the adapters below land first so the switch is a diff about
```
mbegin-robotiq: **Stale in the merged state.** "Nothing links against it yet — the hardware interface still runs the in-tree driver" was true after commit 1, but this same file links `Robotiq::grippers` at line 114 and `hardware_interface.cpp` drives the SDK directly.

Since the branch merges as a unit, whoever reads this next gets told the opposite of what the build does. Worth rewording to describe the end state (the per-commit story is preserved in the commit messages).

grippers/robotiq_driver/CMakeLists.txt
```diff
@@ -26,6 +26,34 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
...
+    "The gripper SDK is missing from ${GRIPPERS_SDK_DIR}. "
+    "Run `git submodule update --init extern/grippers` at the repository root.")
+endif()
+# The SDK declares its library without STATIC or SHARED, so it follows
+# BUILD_SHARED_LIBS. Nothing installs it, so a shared one would leave the
+# plugin needing a library that does not ship. Pinned for this subtree only;
+# the enclosing project keeps its own setting.
+set(BUILD_SHARED_LIBS OFF)
+add_subdirectory(${GRIPPERS_SDK_DIR} ${CMAKE_CURRENT_BINARY_DIR}/grippers_sdk EXCLUDE_FROM_ALL)
+unset(BUILD_SHARED_LIBS)
```
mbegin-robotiq: `unset()` *discards* rather than restores — if the enclosing scope had `BUILD_SHARED_LIBS` as a normal variable, this destroys it instead of putting it back, and the comment above ("the enclosing project keeps its own setting") doesn't hold.

Harmless as things stand: `robotiq_driver` is declared explicitly `SHARED` and the test targets don't care. But it's the same line count to be correct:

```cmake
set(_robotiq_saved_shared "${BUILD_SHARED_LIBS}")
set(BUILD_SHARED_LIBS OFF)
add_subdirectory(...)
set(BUILD_SHARED_LIBS "${_robotiq_saved_shared}")
```

Also on the block just below: `set_target_properties(grippers_sdk ...)` hardcodes the SDK's internal target name while everything else goes through the `Robotiq::grippers` alias. If the SDK renames its target, this line breaks with a less obvious error than the link line would.

grippers/robotiq_driver/CMakeLists.txt
```diff
@@ -68,6 +105,13 @@ target_link_libraries(
   rclcpp::rclcpp
   rclcpp_lifecycle::rclcpp_lifecycle
   serial::serial
+  # $<BUILD_INTERFACE:...> keeps the statically-linked SDK out of the installed
+  # export set, where there would be no target for it to resolve to.
+  # WHOLE_ARCHIVE (CMake 3.24+, satisfied from Jazzy's 24.04 on) because the
+  # SDK's headers ship with this package: a consumer may call any SDK entry
+  # point, so every SDK object has to land in the library — not just the
+  # ones the wrapper itself references.
+  $<BUILD_INTERFACE:$<LINK_LIBRARY:WHOLE_ARCHIVE,Robotiq::grippers>>
```
mbegin-robotiq, line 118: The reasoning is sound given the decision to ship the SDK headers. One consequence worth recording in this comment, because it isn't obvious: `WHOLE_ARCHIVE` means the complete SDK symbol set is exported from a `dlopen`'d pluginlib `.so`.

If a second package in the same process ever links the SDK the same way, you get two copies with independent statics — and `Robotiq::Gripper` owning a serial link and an exchange thread is exactly the kind of thing where that goes wrong quietly. A sentence here would save the next person the archaeology.

grippers/robotiq_driver/CMakeLists.txt
```diff
@@ -111,6 +155,14 @@ install(
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}          # include
 )
 
+# The SDK headers ship alongside ours: they are reachable from this package's
+# public headers, so a consumer including them needs these too.
+install(
+    DIRECTORY ${GRIPPERS_SDK_DIR}/include/
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}          # include
+    FILES_MATCHING PATTERN "*.hpp"                   # not the SDK's own CMakeLists
```
mbegin-robotiq, line 171: `FILES_MATCHING PATTERN "*.hpp"` silently drops anything the SDK ships that isn't `.hpp` — a `.h` C shim, a `.hxx`, or a generated `*_config.h`/`export.h`. The failure lands on a *downstream consumer* at include time, well away from this line, and CI won't catch it because in-tree builds use the `$<BUILD_INTERFACE:>` path where the whole directory is visible.

If the SDK is `.hpp`-only by policy, say so in the comment so it's a checked assumption. Otherwise widen the pattern and exclude `CMakeLists.txt` explicitly instead.

README.md
```diff
@@ -231,6 +231,18 @@ If you already cloned without `--recurse-submodules`:
 git submodule update --init
 ```
 
+Two SDKs live under `extern/` as submodules and are built into the image:
+
+| Submodule | Repository | Used by |
+|---|---|---|
+| `extern/tactile_sensors` | [Robotiq/tactile_sensors](https://github.com/Robotiq/tactile_sensors) | `robotiq_tsf`'s `poll_data_sdk_node` |
+| `extern/grippers` | [Robotiq/grippers](https://github.com/Robotiq/grippers) | the gripper C++ SDK |
```
mbegin-robotiq: The *Used by* cell says "the gripper C++ SDK" — that's what the submodule **is**, not who consumes it. Compare the row above, which correctly names `robotiq_tsf`'s `poll_data_sdk_node`. Should be `robotiq_driver`.

Also a stray double blank line a few lines further down, just before the `| Script | Description |` table.

grippers/robotiq_driver/package.xml
```diff
@@ -19,6 +19,11 @@
   <depend>rclcpp_lifecycle</depend>
   <depend>serial</depend>
 
+  <!-- Serial backend of the gripper SDK compiled in from extern/grippers.
+       No rosdep rule exists for the bare key; installed explicitly in
+       docker/Dockerfile and in CI. -->
+  <depend>libserialport</depend>
```
mbegin-robotiq, line 25: The comment is accurate about Docker and CI (`rosdep-skip-keys: libserialport` is set in the workflow), but it leaves a gap for anyone building this workspace natively: `rosdep install --from-paths src` now fails hard on an unresolvable key, with an error that points at the key rather than at the fix.

One line in the README's build section — install `libserialport-dev`, pass `--skip-keys libserialport` — would close it. Worth doing here since this PR is what makes `robotiq_driver` (not just `robotiq_tsf`) depend on it, so it's newly reachable for gripper-only users.

grippers/robotiq_driver/tests/test_robotiq_gripper_hardware_interface.cpp
```diff
@@ -120,7 +142,111 @@ TEST(TestRobotiqGripperHardwareInterface, exports_expected_command_interfaces)
...
+//! over the SDK's fake gripper, with the exchange slowed down so the recovery
+//! handshake spans many control cycles.
+class RecoveringGripper : public RobotiqGripperHardwareInterface
+{
+public:
+   RecoveringGripper()
+   {
+      parameters_.use_dummy = true;
+      parameters_.closed_position = 0.7929;
+      parameters_.connection.connectionFrequency = 20.0;
```
mbegin-robotiq, line 213: Good test — `deactivation_waits_for_an_in_flight_recovery` exercises a genuinely subtle race rather than a trivial getter, and the rACT reasoning in the docstring is exactly the right thing to have written down.

One problem with the fixture: it sets `parameters_` directly and never runs `on_init`, so **`logger_` is a null `shared_ptr`** when `create_gripper()` hands it to `makeFakeGripper()`. If the SDK doesn't null-check its sink that's a latent crash in the test itself; if it does, the test is still exercising a state the real plugin can't reach (`on_init` always populates it). Either way, one line in the constructor removes the doubt:

```cpp
logger_ = std::make_shared<RclcppLogger>(rclcpp::get_logger("RecoveringGripperTest"));
```

Gaps I'd like to see covered while this file is open, none of them blocking:

- `on_deactivate`'s `kDeactivationTimeout` expiring → `CallbackReturn::ERROR`. Currently untested, and it's the branch that decides whether a gripper is left activated.
- `on_cleanup` with a recovery still in flight (the `recovery_.wait()` path).
- The `write()`-suppressed-during-recovery early return.
- `to_string(Robotiq::ConnectionState)` — cheap, and it's the kind of switch that silently loses a case when the SDK adds an enumerator.

mbegin-robotiq (approved):
Approving — the shape of this migration is right, and extracting the scaling arithmetic and parameter parsing into pure, tested functions is the best part of it. Comments explain *why* rather than *what*, which is rare and welcome.

Inline notes below. Two of them I'd genuinely like fixed before this reaches a real gripper rather than deferred to PR 2:

1. **Zero/NaN denominators in `gripper_scaling.hpp` reach an out-of-range `static_cast<uint8_t>`** — that's UB, and the input is a plausible URDF typo (`gripper_closed_position` of `0`), not a hostile one. The result is an arbitrary `rPR` sent to hardware.
2. **`use_dummy` treats `"False"`, `"0"` and `""` as true** — a capitalisation slip silently runs a simulated gripper in production behind a single INFO line.

Everything else is a one-liner, a comment, or a fair follow-up.

grippers/robotiq_driver/src/hardware_parameters.cpp
ebarnett3: both done

grippers/robotiq_driver/src/hardware_parameters.cpp
ebarnett3: Fatal on non-finite and on `0`, but not on negative:

```cpp
if(!std::isfinite(parameters.closed_position) || parameters.closed_position == 0.0)
```

`<= 0.0` would refuse a description that works today. A joint whose closed
direction is negative is supported — the ratio inverts and the mapping still lands
in the travel band, in PickNik too — so rejecting it would break a working URDF
rather than a broken one. There is a test pinning that.

grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
ebarnett3: The guards are in, but they answer "no register" rather than a substitute value.
Both conversions return `std::optional<uint8_t>`, and `write()` now keeps a
`command_` member so a register it cannot compute this cycle keeps the value it
had; the gripper stays on its previous command and a throttled error names all six
inputs.

The suggested `: kGripperMinPos` would command *fully open* — motion on a gripper
that may be holding a part. A minimum speed or force has the same problem in
milder form: it is still a value nobody asked for, sent to hardware. Since the
position path already skipped the command on a NaN, extending "send nothing"
to the other two registers was the smaller change as well as the safer one.

For the same reason `gripper_max_speed` and `gripper_max_force` are not validated
in `parseParameters` against the `kMaxSpeedDefault` / `kMaxForceDefault`
constants. A description asking for a maximum of zero is a mistake, and silently
substituting the default hides it; the conversion refuses instead and says so.

grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
ebarnett3: see https://github.com/robotiq/ros/pull/20#discussion_r3668847152

grippers/robotiq_driver/src/hardware_interface.cpp
ebarnett3: see https://github.com/robotiq/ros/pull/20#discussion_r3668847152

grippers/robotiq_driver/CMakeLists.txt
ebarnett3: The `BUILD_SHARED_LIBS` half landed as suggested. Naming the target
`Robotiq::grippers` does not work: CMake refuses `set_target_properties` on an
ALIAS target, so the real name stays with a comment saying why.

grippers/robotiq_driver/CMakeLists.txt
ebarnett3: The pattern is unchanged. The SDK ships only `.hpp` — its
`include/Robotiq/CMakeLists.txt` lists nothing else — so widening the glob would
install nothing extra today while implying a rule nobody is checking. The
assumption is stated in a comment instead, where it will read as wrong the day it
becomes wrong.

grippers/robotiq_driver/src/hardware_interface.cpp
ebarnett3: deferred and logged as a Jira ticket. We can also consider using a non-blocking call in the SDK.

grippers/robotiq_driver/src/hardware_interface.cpp
ebarnett3: Jira ticket created for this behavior change

grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
```diff
@@ -0,0 +1,95 @@
...
+#include <optional>
+
+namespace robotiq_driver {
+
+// The usable travel band of the 2F fingers in register counts. The gripper
+// reports and accepts 0..255, but the extremes are outside the mechanical
+// range, so the joint mapping is anchored on these two.
+inline constexpr uint8_t kGripperMinPos = 3;
+inline constexpr uint8_t kGripperMaxPos = 230;
+inline constexpr uint8_t kGripperRange = kGripperMaxPos - kGripperMinPos;
```
l-tschreiber-a11y, line 51: is this the format we want for constants?

grippers/robotiq_driver/src/hardware_parameters.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+#include <cstdint>
+#include <stdexcept>
+#include <string>
+
+#include <rclcpp/logging.hpp>
+
+namespace robotiq_driver {
+namespace {
+constexpr auto kComPortParam = "COM_port";
```
l-tschreiber-a11y: std::string ?

l-tschreiber-a11y (approved):

grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
ebarnett3: I dont have any particular preference for constants. Do you mean using the k prefix for the constants? I think our other convention of using __ is problematic because it conflicts with system constants/macros.

grippers/robotiq_driver/src/hardware_parameters.cpp
ebarnett3: Not `std::string`: a namespace-scope `std::string` cannot be `constexpr` — it
would become a global with dynamic initialization (and an allocation) at
plugin load, and globals with constructors are what the static-init-order
fiasco is made of.

I removed the `auto` , so the type is now spelled out —
`constexpr const char*`. I also changed this for the other constants that were defined this way.

PR #21, opened 2026-07-27
Sgrip 860/retire legacy driver

ebarnett3:
Deletes the Modbus driver PR 1 stopped using:

- the driver
- serial abstraction
- factories
- CRC/data utils
- the fake
-  their tests and mocks
- the vendored `tylerjw/serial` dependency
- both `.repos` files
- the Docker `vcs import` 
- the CI repo-file hook.

The SDK reaches the port through libserialport, which the image and CI already install. 
**An existing workspace keeps a stale `src/serial` checkout that colcon will keep building —
the README migration notes now say to delete it.** No behaviour change:
nothing removed was reachable at the tip of PR 1.

The twelve headers this package used to install stay behind as signposts that
fail with an `#error` naming the replacement, rather than a missing-file
error. The stubs say "will be deleted in a future
release"; picking that release is a follow-up.

The README migration instructions are rewritten for the submodule world, and
two new sections document the full hardware-parameter table and the
activation/fault/recovery semantics — in particular that **activating or
deactivating runs the reset handshake, releasing any grip**. That has always
been true of this driver and was previously undocumented.

A second commit folds the last standalone test executable into
`robotiq_driver_tests`: one gmock binary per package, 25 cases unchanged,
`--gtest_filter` picks a subset.


grippers/robotiq_driver/include/robotiq_driver/fake/fake_driver.hpp
```diff
@@ -26,44 +26,17 @@
...
-   bool activated_ = false;
-   uint8_t position_ = 0;
-   bool gripper_is_moving_ = false;
-   uint8_t speed_ = 0;
-   uint8_t force_ = 0;
-};
+#pragma once
 
-} // namespace robotiq_driver
+#error "robotiq_driver/fake/fake_driver.hpp was removed: use makeFakeGripper() from <Robotiq/fake/gripper_factory.hpp>."
```
mbegin-robotiq: 👍

grippers/robotiq_driver/include/robotiq_driver/default_driver.hpp
```diff
@@ -26,99 +26,16 @@
...
-
-   /**
-    * @brief Commands the gripper to move to the desired position.
-    * @param pos A value between 0x00 (fully open) and 0xFF (fully closed).
-    */
-   void set_gripper_position(uint8_t pos) override;
+//! \brief Removed header, kept as a signpost.
+//! This package no longer implements its own gripper driver: the transport,
+//! the register map and the runtime API all come from the Robotiq gripper SDK
+//! (extern/grippers). The header you included went with that change.
```
mbegin-robotiq, line 32: do we really need to keep all these signposts ?

mbegin-robotiq (approved):

grippers/robotiq_driver/include/robotiq_driver/default_driver.hpp
ebarnett3: These files were accessible by users of the PickNik package so if we just delete them, it will cause "file missing" compilation errors for users that included them. The signposts guide users that are transitioning to the Robotiq package. I think we can delete them after a release or two.

l-tschreiber-a11y (approved):

PR #24, opened 2026-07-31
Support humble lyrical rolling

ebarnett3:
## Summary

Supports ROS 2 Humble, Jazzy and Lyrical — the three live LTS distros — from this one branch, plus Rolling built non-blocking as early warning. `main` previously built on Jazzy only.

On Humble this is a drop-in replacement for PickNik's `humble` branch: same controller, same `control_msgs/GripperCommand` action, same launch files, and a repos file with identical contents, so existing action clients keep working untouched.

> **Depends on [Robotiq/grippers#13](https://github.com/Robotiq/grippers/pull/13).** `extern/grippers` is pinned to a commit on that PR's branch so CI could exercise the pair. **The pin must be moved to the merged SHA on the SDK's `main` before this merges** — a squash-merge there changes the SHA. **To prevent an accidental merge, this PR is left as a draft until the pin has been moved.**

## Problem

The two new distros broke in disjoint places — Lyrical in CMake, Humble in the `ros2_control` API — and the vendored gripper SDK assumed Jazzy throughout:

| Package | Humble | Jazzy | Lyrical |
|---|---|---|---|
| `robotiq_tsf` | builds | builds | **fails configure** |
| `robotiq_driver` / `robotiq_controllers` | **fail compile** | build | build |
| `robotiq_description` | `rosdep` unresolvable | builds | builds |

Lyrical's `ament_cmake` removed `ament_target_dependencies()`, and its CMake 4 wants `C` enabled before the SDK subproject asks. Humble's `ros2_control` 2.x passes `HardwareInfo` to `on_init()`, its command handles have no `get_optional()` or bool-returning `set_value()`, its GCC 11 doesn't supply `<cmath>` and friends transitively, and its CMake 3.22 has no `$<LINK_LIBRARY:WHOLE_ARCHIVE>`. And `parallel_gripper_controller` does not exist on Humble — ROS only gained it in Jazzy.

## Changes

- `robotiq_tsf`: imported targets instead of `ament_target_dependencies()`; fix `rclpy`'s removed `warn()` and the deprecated `spin_some(node)`.
- `robotiq_driver` / `robotiq_controllers`: a compat header per package, detecting each API by shape (`__has_include`, `std::void_t` on the handle) rather than a version threshold, since Jazzy's 4.x line keeps moving. The Humble path is byte-for-byte PickNik's. Adds `robotiq_controllers`' first tests, covering both handle shapes.
- `robotiq_driver` links the SDK's new object library, `Robotiq::grippers_objects`, so every object lands in the plugin with no whole-archive flags and no CMake version floor.
- `robotiq_description`: a Humble controller config using stock `position_controllers/GripperActionController`, selected on `$ROS_DISTRO`, and `condition=` deps so `rosdep` resolves on every distro with no `--skip-keys`.
- `use_fake_hardware` now behaves like hardware everywhere: the mock gets the gripper's effort/velocity command interfaces, and no longer owns the mimicked finger joints, so `robot_state_publisher` derives them as it does on hardware.
- CI matrix over the LTS distros inside `ros:<distro>-ros-base` containers; `docker/run.sh` takes `ROS_DISTRO`.

## Testing

CI is green on all four distros. Locally, all six packages build with no warnings from our code and `rosdep` resolves clean.

Hardware, real 2F-85 and TSF-85: 7 sensor topics including fused orientation; `full_test` from the installed path; 3 controllers active with `joint_states` at 500 Hz; `gripper_cmd` goals driving the gripper to target within 0.001 rad; and `~/reactivate_gripper` returning success — the only path that exercises the handle compat shim against real hardware. Re-run after the SDK bump, including holding and motion phases with no position hunting.

## Notes

- **The `gripper_cmd` action type differs by distro**: `GripperCommand` on Humble, `ParallelGripperCommand` on Jazzy and Lyrical. That break is upstream ROS's, not ours, and it is what lets Humble users migrate with no client changes. README carries the table.
- Three pre-existing bugs fixed along the way: `full_test` was built but never installed, so `ros2 run` could never find it; a dead `robotiq_update_rate.yaml` reference warned on every bringup; and controller parameters stopped reaching controllers on Lyrical because `controller_manager` no longer forwards its parameter file — now passed with the spawner's `--param-file`.
- **Deprecation debt**: the driver overrides `export_state_interfaces()` / `export_command_interfaces()`, which `ros2_control` 6.x marks deprecated in favour of `on_export_state_interfaces()`. Nothing warns about it today, so the non-blocking Rolling job is how we will see the break coming.
- 
FYI @JonathanBeaudoinRq @jlongvalRobotiq @bcastets-robotiq @jkski-robotiq 

### DIFF
--- a/grippers/robotiq_controllers/test/test_ros2_control_compat.cpp
+++ b/grippers/robotiq_controllers/test/test_ros2_control_compat.cpp
@@ -1,0 +1,118 @@
+// Copyright (c) 2026 Robotiq
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//    * Redistributions of source code must retain the above copyright
+//      notice, this list of conditions and the following disclaimer.
+//
+//    * Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//    * Neither the name of the copyright holder nor the names of its
+//      contributors may be used to endorse or promote products derived from
+//      this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+// The activation controller only ever compiles against the handle API of the
+// distro it is built on, so the branch that distro does not use would otherwise
+// go untested. These fakes stand in for both LoanedCommandInterface shapes so
+// that every distro exercises both.
+
+#include <gtest/gtest.h>
+
+#include <optional>
+
+#include <robotiq_controllers/ros2_control_compat.hpp>
+
+namespace robotiq_controllers::test {
+
+/// Humble (ros2_control 2.x): writes cannot fail, reads always yield a value.
+class OldApiHandle
+{
+public:
+   void set_value(double value) { value_ = value; }
+   double get_value() const { return value_; }
+
+private:
+   double value_ = 0.0;
+};
+
+/// Jazzy 4.x and later: writes report success, reads may come back empty.
+class NewApiHandle
+{
+public:
+   bool set_value(double value)
+   {
+      if(!writable_)
+      {
+         return false;
+      }
+      value_ = value;
+      return true;
+   }
+
+   std::optional<double> get_optional() const { return readable_ ? std::optional<double>{value_} : std::nullopt; }
+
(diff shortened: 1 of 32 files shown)


grippers/robotiq_controllers/test/test_ros2_control_compat.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+/// Humble (ros2_control 2.x): writes cannot fail, reads always yield a value.
+class OldApiHandle
+{
+public:
+   void set_value(double value) { value_ = value; }
+   double get_value() const { return value_; }
+
+private:
+   double value_ = 0.0;
```
l-tschreiber-a11y, line 50: value_   ->  _value  ?

grippers/robotiq_controllers/test/test_ros2_control_compat.cpp
```diff
@@ -0,0 +1,118 @@
...
+};
+
+/// Jazzy 4.x and later: writes report success, reads may come back empty.
+class NewApiHandle
+{
+public:
+   bool set_value(double value)
```
l-tschreiber-a11y, line 57: camelCase ?

l-tschreiber-a11y (approved):

mbegin-robotiq (approved):

grippers/robotiq_controllers/test/test_ros2_control_compat.cpp
ebarnett3: `set_value` and the other methods here aren't ours to choose — they're the method names on the real
ros2_control handles these fakes stand in for. From the Jazzy headers:

```
loaned_command_interface.hpp:117   [[nodiscard]] bool set_value(const T&, unsigned int max_tries = 10)
loaned_command_interface.hpp:138   double get_value() const     // deprecated in favour of get_optional()
handle.hpp:329                     [[nodiscard]] std::optional<T> get_optional() const
```

`ros2_control_compat.hpp` dispatches on those names literally — `HasGetOptional`
keys off `decltype(...get_optional())` and `SetValueReturnsBool` off
`decltype(...set_value(0.0))` — and at runtime it is handed real handles. So the
fakes have to expose the same spelling to be substitutable; renaming them would
mean renaming the calls in the header too, and then it no longer compiles against
ros2_control.

grippers/robotiq_controllers/test/test_ros2_control_compat.cpp
ebarnett3: Good instinct, and it's the gripper SDK's convention (`_mutex`, `_clock`,
`_value`), but this repo goes the other way — trailing underscore, without
exception:

| | trailing `value_` | leading `_value` |
|---|---|---|
| this repo, hand-written sources | `msg_` ×60, `logger_` ×7, `gripper_position_` ×6, `value_` ×3, … | **0** |
| grippers SDK | — | `_clock`, `_last`, `_mutex`, `_period`, `_value`, `_what` |

So changing it here would make this file the only outlier. I'd rather not do it
in this PR and instead align the whole repo in one pass — worth covering method
names in the same go, since those aren't fully consistent either (`info_with` in
`test_hardware_parameters.cpp` is snake_case where our own code is otherwise
camelCase).

Happy to open a ticket for that pass if you agree.

PR #39, opened 2026-08-15
feat(robotiq_driver): expose object detection and effort as state interfaces

### LINKED ISSUE #26: robotiq_driver: expose effort (gCU) and object-detection (gOBJ) state interfaces
The `ros2_control` hardware interface exports only `position` and `velocity` state interfaces. Neither object detection (`gOBJ`) nor motor current (`gCU`, the standard grip-force proxy) is exposed, so downstream controllers cannot do force-controlled grasping, compliance, contact detection, or even tell whether the gripper caught the part.

Carried over from PickNik: [PickNikRobotics/ros2_robotiq_gripper#118](https://github.com/PickNikRobotics/ros2_robotiq_gripper/issues/118).

## Reproduce

```bash
ros2 launch robotiq_description robotiq_control.launch.py com_port:=/dev/ttyUSB0
# second shell
ros2 control list_hardware_interfaces
```

Observed:

```
command interfaces
    reactivate_gripper/reactivate_gripper_cmd [available] [claimed]
    reactivate_gripper/reactivate_gripper_response [available] [claimed]
    robotiq_85_left_knuckle_joint/position [available] [claimed]
    robotiq_85_left_knuckle_joint/set_gripper_max_effort [available] [unclaimed]
    robotiq_85_left_knuckle_joint/set_gripper_max_velocity [available] [unclaimed]
state interfaces
    robotiq_85_left_knuckle_joint/position
    robotiq_85_left_knuckle_joint/velocity
```

## Analysis

Present today: `position` (state + command), `velocity` (state, derived from the position derivative rather than read from hardware), `set_gripper_max_effort` / `set_gripper_max_velocity` (commands that set `rFR` / `rSP` for the firmware motion profile — commands, not feedback), and `reactivate_gripper_*`.

Missing:

- **`effort` state interface**, fed from `gCU` (motor current) — the standard grip-force proxy on the 2F-85.
- **Object-detection state interface**, fed from `gOBJ` (2 bits: 0 = moving, 1 = object held while opening, 2 = object held while closing, 3 = position reached). Without it there is no way to distinguish a successful grasp from a missed one, which is a primitive any pick-and-place needs.

Both values already arrive in the status block the SDK exchanges every cycle, so this is a matter of exporting them.

## Target

- a `<joint>/effort` state interface fed by `gCU`
- a state interface for `gOBJ` (naming TBD — `<joint>/object_status` or similar)

ebarnett3:
Closes #26.

The hardware interface exported position and velocity only, so nothing downstream could tell a caught part from a missed one, or judge a grip. Both values already arrive in the status block the SDK exchanges every cycle.

Export effort from gCU and object_status from gOBJ. Effort is read back through the same full scale the rFR command is written against, so the effort reported and the effort requested share a scale; it stays a motor-current proxy, not a measurement of the grip. object_status carries gOBJ unchanged — 0 moving, 1 held while opening, 2 held while closing, 3 at the requested position — since a remapping would only cost consumers the manual's vocabulary.

on_init no longer demands exactly two state interfaces. A description may now declare any of the four, and is rejected only for naming a fifth the hardware cannot answer for.

Measured on a 2F-85 over a reactivation sweep: object_status ran 3 -> 0 -> 3, effort ran 0 to 92.16 N in steps of 0.9216 N, which is the 235 N full scale over the 255 counts of the register.

## Verification

- 40 `robotiq_driver` tests pass, including three new ones: two on the register/SI round trip and one pinning the exported state-interface names.
- Fake-hardware bringup (`use_fake_hardware:=true`) still activates all three controllers with the two new interfaces declared.
- On a real 2F-85, `ros2 control list_hardware_interfaces` reports `effort` and `object_status`, and both carry live values on `/dynamic_joint_states`.

### DIFF
--- a/grippers/robotiq_description/urdf/2f_85.ros2_control.xacro
+++ b/grippers/robotiq_description/urdf/2f_85.ros2_control.xacro
@@ -69,13 +69,10 @@
                     <param name="initial_value">${gripper_closed_position}</param>
                 </state_interface>
                 <state_interface name="velocity"/>
-                <!-- Fed from the status block the driver already exchanges:
-                     effort from gCU (a motor-current proxy, on the same full
-                     scale as the force command), object_status from gOBJ
-                     (0 moving, 1 held while opening, 2 held while closing,
-                     3 at the requested position). -->
-                <state_interface name="effort"/>
-                <state_interface name="object_status"/>
+                <xacro:unless value="${use_fake_hardware or sim_gazebo or sim_isaac}">
+                    <state_interface name="motor_current"/>
+                    <state_interface name="object_status"/>
+                </xacro:unless>
             </joint>
             <!-- When simulating we need to include the rest of the gripper joints -->
             <xacro:if value="${sim_isaac or sim_gazebo}">
--- a/grippers/robotiq_description/urdf/2f_85.ros2_control.xacro
+++ b/grippers/robotiq_description/urdf/2f_85.ros2_control.xacro
@@ -69,6 +69,13 @@
                     <param name="initial_value">${gripper_closed_position}</param>
                 </state_interface>
                 <state_interface name="velocity"/>
+                <!-- Fed from the status block the driver already exchanges:
+                     effort from gCU (a motor-current proxy, on the same full
+                     scale as the force command), object_status from gOBJ
+                     (0 moving, 1 held while opening, 2 held while closing,
+                     3 at the requested position). -->
+                <state_interface name="effort"/>
+                <state_interface name="object_status"/>
             </joint>
             <!-- When simulating we need to include the rest of the gripper joints -->
             <xacro:if value="${sim_isaac or sim_gazebo}">
--- a/grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
+++ b/grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
@@ -92,4 +92,11 @@ inline constexpr uint8_t kGripperRange = kGripperMaxPos - kGripperMinPos;
    }
    return static_cast<uint8_t>(std::min(value / maximum, 1.0) * 0xFF);
 }
+
+//! The inverse: a 0x00..0xFF register read back as the fraction of \p maximum
+//! it stands for.
+[[nodiscard]] inline double fractionOfFromRegister(uint8_t register_value, double maximum)
+{
+   return maximum * register_value / 0xFF;
+}
 } // namespace robotiq_driver
--- a/grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
+++ b/grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
@@ -93,10 +93,10 @@ inline constexpr uint8_t kGripperRange = kGripperMaxPos - kGripperMinPos;
    return static_cast<uint8_t>(std::min(value / maximum, 1.0) * 0xFF);
 }
 
-//! The inverse: a 0x00..0xFF register read back as the fraction of \p maximum
-//! it stands for.
-[[nodiscard]] inline double fractionOfFromRegister(uint8_t register_value, double maximum)
+inline constexpr double kAmperesPerCurrentCount = 0.010;
+
+[[nodiscard]] inline constexpr double motorCurrentFromRegister(uint8_t register_value)
 {
-   return maximum * register_value / 0xFF;
+   return kAmperesPerCurrentCount * register_value;
 }
 } // namespace robotiq_driver
--- a/grippers/robotiq_driver/include/robotiq_driver/hardware_interface.hpp
+++ b/grippers/robotiq_driver/include/robotiq_driver/hardware_interface.hpp
@@ -58,10 +58,9 @@
 #include <rclcpp/rclcpp.hpp>
 
 namespace robotiq_driver {
-//! Name of the state interface carrying gOBJ. ros2_control defines HW_IF_
-//! constants for position, velocity and effort but has nothing for object
-//! detection, so the URDF and every consumer spell this one out.
+
 inline constexpr const char* kObjectStatusInterface = "object_status";
+inline constexpr const char* kMotorCurrentInterface = "motor_current";
 
 //! ros2_control hardware interface for a Robotiq 2F gripper, driven through the
 //! gripper SDK.
@@ -119,8 +118,8 @@ public:
    CallbackReturn on_error(const rclcpp_lifecycle::State& previous_state) override;
 
    /**
-    * This method exposes position, velocity, effort and object status of joints
-    * for reading.
+    * This method exposes position, velocity, motor current and object status of
+    * joints for reading.
     */
    ROBOTIQ_DRIVER_PUBLIC
    std::vector<hardware_interface::StateInterface> export_state_interfaces() override;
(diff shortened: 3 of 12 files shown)


grippers/robotiq_driver/include/robotiq_driver/hardware_interface.hpp
```diff
@@ -179,6 +185,17 @@ class RobotiqGripperHardwareInterface : public hardware_interface::SystemInterfa
...
+
+   // gCU read back through the same full-scale force the rFR command uses, so
+   // the effort reported and the effort requested are on one scale. It is a
+   // motor-current proxy either way, not a measurement of the grip.
+   double gripper_effort_ = 0.0;
+
+   // gOBJ as it comes off the wire: 0 moving, 1 held while opening, 2 held
+   // while closing, 3 at the requested position. A double because that is what
+   // a state interface carries; consumers compare it against those four values.
+   double gripper_object_status_ = 0.0;
```
mbegin-robotiq: These two default to `0.0`, but unlike `gripper_position_`/`gripper_velocity_`/the command they are not overwritten with NaN in `on_init`, and not seeded in `on_activate` next to `gripper_position_`.

For `object_status` that matters: `0.0` is not a neutral placeholder, it is the meaningful value gOBJ = 0, "moving". Between activation and the first successful `read()` the interface therefore reports "moving" while the fingers may be sitting settled at gOBJ = 3, and a consumer has no NaN to test for "no reading yet".

Suggest NaN-initializing both alongside the existing members in `on_init` and seeding them in `on_activate` from the first status exchange.

grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
```diff
@@ -92,4 +92,12 @@ inline constexpr uint8_t kGripperRange = kGripperMaxPos - kGripperMinPos;
...
    return static_cast<uint8_t>(std::min(value / maximum, 1.0) * 0xFF);
 }
+
+//! The inverse: a 0x00..0xFF register read back as the fraction of \p maximum
+//! it stands for. Total where registerFromFractionOf is not — every byte names
+//! a fraction, so there is no reading the arithmetic cannot map.
+[[nodiscard]] inline double fractionOfFromRegister(uint8_t register_value, double maximum)
+{
+   return maximum * register_value / 0xFF;
+}
```
mbegin-robotiq: This validates nothing, while its inverse `registerFromFractionOf` rejects non-positive and non-finite `maximum` — and `gripper_max_force` is not validated at parse time either (`hardware_parameters.cpp`).

That asymmetry changes a failure mode: before this PR a negative or `nan` `gripper_max_force` only made `write()` fail (throttled error, previous command kept, state untouched). Now the same bad parameter silently poisons the published `effort` state every read cycle — `nan` or a negative newton value, with no error anywhere.

Either mirror the guard here, or better, validate `gripper_max_force` once at parse time so both directions can assume it is finite and positive.

grippers/robotiq_description/urdf/2f_85.ros2_control.xacro
```diff
@@ -69,6 +69,13 @@
                     <param name="initial_value">${gripper_closed_position}</param>
                 </state_interface>
                 <state_interface name="velocity"/>
+                <!-- Fed from the status block the driver already exchanges:
+                     effort from gCU (a motor-current proxy, on the same full
+                     scale as the force command), object_status from gOBJ
+                     (0 moving, 1 held while opening, 2 held while closing,
+                     3 at the requested position). -->
+                <state_interface name="effort"/>
+                <state_interface name="object_status"/>
```
mbegin-robotiq: These are declared unconditionally, outside any `sim_*` / `use_fake_hardware` guard, so they land on the mock, gazebo and isaac plugin branches too (same in `2f_140.ros2_control.xacro:66`).

Under `use_fake_hardware`, `GenericSystem` will happily export `object_status` but nothing ever writes it, so it stays pinned at `0.0` = "moving". A client that waits for 1/2/3 to confirm a grasp works on hardware and hangs forever in simulation. `gz_ros2_control` does not register the non-standard `object_status` name at all, so that branch differs again.

The PR notes verify that fake bringup *activates* — worth also checking that the value moves, and if it cannot, guarding these two to the real-hardware branch so the difference is explicit rather than a silent lie.

mbegin-robotiq (commented):
Three inline notes on the new state interfaces — all small, none blocking.

mbegin-robotiq (approved):
Approving — the three inline threads are small fixes, not blockers.

One thing left open deliberately and not filed as a thread: the `effort` interface publishes gCU scaled by `gripper_max_force`, a command-side limit, so the same physical motor current reads 92.2 N with `gripper_max_force:=235` and 39.2 N with `:=100`. It is neither the current nor the grip force, and a consumer thresholding it has to know the URDF. That is a design call rather than a defect in this diff, so it is your call whether to rename the semantics or document them.

grippers/robotiq_driver/include/robotiq_driver/hardware_interface.hpp
ebarnett3: Fixed. Both start as `quiet_NaN()` in `on_init` alongside the existing members.

grippers/robotiq_description/urdf/2f_85.ros2_control.xacro
ebarnett3: Confirmed and fixed in eab84fb. I reproduced it before changing anything — under `use_fake_hardware:=true`, `/dynamic_joint_states` read:

```
interface_names: [effort, position, object_status, velocity]
values:          [0.0,    0.7929,   0.0,           0.0]
```

`position` at 0.7929 means fully closed and settled, while `object_status` reported 0, "moving". Exactly the silent lie you predicted.

Both interfaces are now guarded to the real-hardware branch, as you suggested, in `2f_85` and `2f_140`. A consumer in simulation fails to find the interface instead of reading a plausible wrong number.

grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp
ebarnett3: Fixed, though not by mirroring the guard — the function is gone. effort no longer scales gCU by gripper_max_force at all (see the reply on the effort semantics below), so there is no second direction left to validate.

ebarnett3:
> Approving — the three inline threads are small fixes, not blockers.
> 
> One thing left open deliberately and not filed as a thread: the `effort` interface publishes gCU scaled by `gripper_max_force`, a command-side limit, so the same physical motor current reads 92.2 N with `gripper_max_force:=235` and 39.2 N with `:=100`. It is neither the current nor the grip force, and a consumer thresholding it has to know the URDF. That is a design call rather than a defect in this diff, so it is your call whether to rename the semantics or document them.

Taking this one as blocking rather than as a follow-up, because it is the only item here that gets more expensive after release: once people threshold against a published value, the scale is frozen.

I went to the manual, and it settles it in both directions.

It gives a conversion, but not the one we would want. §*Register: CURRENT* defines gCU as "approximate current equivalent is 10 * value read in mA" — a real physical quantity, fixed by the hardware, independent of anything in the description. It is not in the SDK, which documents the field only as "gCU (effort proxy)".

It rules out a force conversion. The force depends on many factors, so approximating it with the current will often be very inaccurate. We therefore publish the documented quantity and drops the invented one: `effort` becomes `motor_current`, in amperes, `gCU * 0.010`, with no scale parameter in the signature to get wrong.

PR #41, opened 2026-08-27
Fix the rosdistro key for libserialport

IamPhytan:


docker/Dockerfile
ebarnett3: stale comment: libserialport should be removed

robotiq_tsf/package.xml
```diff
@@ -31,8 +31,8 @@
   <!-- Quaternion algebra in the AHRS/algorithms library (header-only). -->
   <depend>eigen</depend>
   <!-- SDK-backed poller (poll_data_sdk_node) links the tactile_sensors SDK,
-       which uses libserialport. Installed explicitly in docker/Dockerfile. -->
-  <depend>libserialport</depend>
+       which uses libserialport. -->
+  <depend>libserialport-dev</depend>
```
ebarnett3: `<depend>` makes `libserialport-dev` a *build, build_export and exec* dependency. Concrete consequence: `rosdep install --dependency-types exec` on a deployment host — and any bloom-released binary deb of these packages — will now pull `libserialport-dev` (headers plus the `.so` dev symlink) rather than the runtime `libserialport0`. That directly contradicts `docker/Dockerfile`'s runtime stage, which deliberately installs `libserialport0` and carries a comment complaining about `-dev` packages leaking in (~350 MB). Since rosdep has no runtime-only `libserialport` key, the accurate declaration is `<build_depend>libserialport-dev</build_depend>` + `<build_export_depend>libserialport-dev</build_export_depend>`, with the runtime library handled the way the runtime stage already does.

ebarnett3 (approved):
Thanks for the contribution!

docker/Dockerfile
IamPhytan: Fixed in last commit

robotiq_tsf/package.xml
IamPhytan: Fix in new commit

ebarnett3: Apologies — bad advice on my part, reverted in a commit I pushed to your branch.

The runtime stage never runs rosdep (`ros-core` + explicit apt installs), so exec deps don't affect the image, and `-dev` is only 184 KB vs 81 KB. Meanwhile `poll_data_sdk_node` links `libserialport.so.0`, so without the exec dep a released deb declares no runtime dependency on it. Plain `<depend>`, same as `robotiq_driver`.

Pushed it directly so we can merge today — say if you'd rather have done it yourself. Rest of the PR is spot on.

PR #43, opened 2026-08-31
Issue 35/baudrate launch arg

### LINKED ISSUE #35: Add baudrate as a configurable URDF / launch arg
The URDF hardcodes the Modbus baudrate at 115200. The driver's `hardware_interface` already reads a `baudrate` hardware parameter, but the URDF/xacro does not expose one, so it cannot be overridden from the launch file — a gripper configured for a different baudrate cannot be driven without editing the description.

## Change

- Add a `baudrate` xacro arg and a matching `<param>` in the `ros2_control` hardware block, for both the 2F-85 and the 2F-140.
- Add a matching `DeclareLaunchArgument` in `robotiq_control.launch.py`, defaulting to the current 115200 so nothing changes for existing users.

ebarnett3:
Closes #35.

The baudrate was reachable only by editing the xacro. It is now a launch argument
on `robotiq_control.launch.py`, threaded through the top-level descriptions, the
gripper macros and the `ros2_control` blocks for both the 2F-85 and the 2F-140.
`gripper_tactile_viz.launch.py` forwards it too, so the tactile path is not stuck
at 115200.

Because the value is now user-facing, the driver validates it. `std::stoul` alone
ignores trailing characters, so `115200bps` parsed as 115200 and `1.152e5` as 1,
and `0` or a wrapped `-1` reached `sp_set_baudrate` unchallenged — each leaving a
port that opens and then answers nothing. Parsing now rejects trailing characters
and bounds the rate, falling back to the default with a log line as the other
malformed parameters do. `slave_address` had the same latent bug and shares the fix.

## Testing

Verified on a 2F-85, reconfiguring the gripper itself between rates:

| gripper | `baudrate:=115200` | `baudrate:=57600` |
|---|---|---|
| 115200 | connects, activates | fails in `on_configure` |
| 57600  | fails in `on_configure` | connects, activates |

A mismatch fails before activation, so a wrong rate cannot move the gripper. New
unit tests cover the rejected spellings; a new xacro test asserts the argument
reaches `<param name="baudrate">`, which is the half-updated-macro regression.

### DIFF
--- a/grippers/robotiq_driver/src/hardware_parameters.cpp
+++ b/grippers/robotiq_driver/src/hardware_parameters.cpp
@@ -90,6 +90,32 @@ T parameterOr(const hardware_interface::HardwareInfo& info,
    }
 }
 
+//! Parse \p text in \p base, rejecting the trailing characters std::stoull
+//! silently ignores: "115200bps" would otherwise read as 115200 and "1.152e5"
+//! as 1.
+uint64_t asWholeNumber(const std::string& text, int base = 10)
+{
+   std::size_t consumed = 0;
+   const uint64_t value = std::stoull(text, &consumed, base);
+   if(consumed != text.size())
+   {
+      throw std::invalid_argument("expected a whole number");
+   }
+   return value;
+}
+
+//! The range libserialport accepts. Out of it the port still opens and every
+//! exchange then times out, which reads as a dead gripper rather than a typo.
+uint32_t asBaudrate(const std::string& text)
+{
+   const uint64_t value = asWholeNumber(text);
+   if(value == 0 || value > 1000000)
+   {
+      throw std::out_of_range("baudrate must be between 1 and 1000000");
+   }
+   return static_cast<uint32_t>(value);
+}
+
 double asDouble(const std::string& text)
 {
    return std::stod(text);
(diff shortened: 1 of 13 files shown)


mbegin-robotiq (approved):

grippers/robotiq_driver/src/hardware_parameters.cpp
```diff
@@ -90,6 +90,32 @@ T parameterOr(const hardware_interface::HardwareInfo& info,
...
+
+//! The range libserialport accepts. Out of it the port still opens and every
+//! exchange then times out, which reads as a dead gripper rather than a typo.
+uint32_t asBaudrate(const std::string& text)
+{
+   const uint64_t value = asWholeNumber(text);
+   if(value == 0 || value > 1000000)
+   {
+      throw std::out_of_range("baudrate must be between 1 and 1000000");
+   }
```
jlongvalRobotiq, line 115: in reality, should'nt we accept only a baudrate from a set of acceptable baudrates like 9600, 19200, 38400, 57600 and 115200 ?

ebarnett3: Maybe, but in the Robotiq User Interface there is no such restriction, and I have verified (by mistake) in the past that nonstandard baudrates work.

PR #47, opened 2026-09-02
feat(robotiq_description): let the sim_isaac / sim_gazebo plugins activate from the launch

jlongvalRobotiq:
## Problem

The gripper xacros have let you pick `sim_isaac` (topic_based_ros2_control) or `sim_gazebo` (gz_ros2_control) as the hardware plugin for years, but the controller config and the launch file were written for the real driver only. Pick a sim plugin and `robotiq_gripper_controller` never activates: the config claims `set_gripper_max_effort` / `set_gripper_max_velocity`, which only the driver exports, and the launch spawns `robotiq_activation_controller`, whose GPIO the sim URDF does not declare. The launch also never exposed the sim flags at all.

## Change

One commit per layer, oldest first:

1. **xacro** — the top-level 2F-85 / 2F-140 xacros accept and forward `sim_isaac`, `sim_gazebo`, `isaac_joint_commands`, `isaac_joint_states` (the macro already had them).
2. **config** — `robotiq_controllers.sim.yaml` (+ Humble variant): same controllers as the hardware config minus the activation controller and the two driver-only command interfaces.
3. **launch** — new arguments, defaulting to today's behaviour; either sim flag selects the sim config and skips the activation spawner.
4. **README** — the two sim launch lines and a note that `TopicBasedSystem` matches joints by name.

Hardware and mock bringup are unchanged: same files, same controllers, same action surface. Neither sim plugin becomes a dependency.

## Verification

- 24 → 54 tests in `robotiq_description`; green at every commit. New tests pin what each sim plugin exports, that the sim config claims only that, and the launch's own spawner wiring per distro.
- Jazzy: `use_fake_hardware:=true` still activates all three controllers; `sim_isaac:=true` exercised live against an Isaac Sim 5.1 twin of a 2F-85 (open, close, stall-on-object through `gripper_cmd`).
- black, flake8, codespell clean.

## Note
only the last 4 commits constitute this PR. The commits before that come from PRs that havent yet been merged #48 and #49 )

### DIFF
--- a/grippers/robotiq_description/urdf/robotiq_2f_85_gripper.urdf.xacro
+++ b/grippers/robotiq_description/urdf/robotiq_2f_85_gripper.urdf.xacro
@@ -1,15 +1,5 @@
 <?xml version="1.0"?>
 <robot xmlns:xacro="http://wiki.ros.org/xacro" name="robotiq_gripper">
-    <!-- parameters -->
-    <xacro:arg name="use_fake_hardware" default="true" />
-    <xacro:arg name="com_port" default="/dev/ttyUSB0" />
-    <xacro:arg name="baudrate" default="115200" />
-
-    <!-- Import macros -->
-    <xacro:include filename="$(find robotiq_description)/urdf/robotiq_2f_85_macro.urdf.xacro" />
-
-    <link name="world" />
-    <xacro:robotiq_gripper name="RobotiqGripperHardwareInterface" prefix="" parent="world" use_fake_hardware="$(arg use_fake_hardware)" com_port="$(arg com_port)" baudrate="$(arg baudrate)">
-        <origin xyz="0 0 0" rpy="0 0 0" />
-    </xacro:robotiq_gripper>
+    <xacro:arg name="model" default="2f_85" />
+    <xacro:include filename="$(find robotiq_description)/urdf/robotiq_2f_gripper.urdf.xacro" />
 </robot>
--- a/README.md
+++ b/README.md
@@ -133,8 +133,11 @@ In the combined launch the pad frames are TF-mounted on the gripper fingertip li
 ROS 2 `ros2_control` driver for Robotiq 2F adaptive grippers, under [`grippers/`](grippers/).
 
 Descriptions ship for the **2F-85** and the **2F-140**; `robotiq_control.launch.py` defaults to the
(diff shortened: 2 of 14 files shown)


grippers/robotiq_description/urdf/robotiq_2f_85_gripper.urdf.xacro
```diff
@@ -4,12 +4,18 @@
...
+    <xacro:arg name="isaac_joint_states" default="/isaac_joint_states" />
 
     <!-- Import macros -->
     <xacro:include filename="$(find robotiq_description)/urdf/robotiq_2f_85_macro.urdf.xacro" />
 
     <link name="world" />
-    <xacro:robotiq_gripper name="RobotiqGripperHardwareInterface" prefix="" parent="world" use_fake_hardware="$(arg use_fake_hardware)" com_port="$(arg com_port)" baudrate="$(arg baudrate)">
+    <xacro:robotiq_gripper name="RobotiqGripperHardwareInterface" prefix="" parent="world" use_fake_hardware="$(arg use_fake_hardware)" com_port="$(arg com_port)" baudrate="$(arg baudrate)"
+        sim_isaac="$(arg sim_isaac)" sim_gazebo="$(arg sim_gazebo)"
+        isaac_joint_commands="$(arg isaac_joint_commands)" isaac_joint_states="$(arg isaac_joint_states)">
```
jlongvalRobotiq: lots of repetition with the 2f 140 file.

jlongvalRobotiq: Fixed in ff246ab. The two top-level files were identical apart from the macro they include, so the body now lives in `robotiq_2f_gripper.urdf.xacro` keyed by a `model` xacro arg (`2f_85` / `2f_140`), and `robotiq_2f_85_gripper.urdf.xacro` / `robotiq_2f_140_gripper.urdf.xacro` are 4-line wrappers that set the default model and include it. Existing `model:=.../robotiq_2f_140_gripper.urdf.xacro` launch paths, the tests, and the README are unchanged.

Verified the expanded URDF is byte-identical (minus xacro's "autogenerated from" header) for both models under mock, driver, sim_isaac and sim_gazebo; `test_ros2_control_urdf.py` passes.

jlongvalRobotiq:
Superseded by #52, same commits, opened as an in-repo branch so the stack (#50 → #51 → #52) shows only each PR's own commits.

PR #48, opened 2026-09-02
refactor(robotiq_description): share the top-level and ros2_control xacros between the 2F models

jlongvalRobotiq:
## Summary
- The 2F-85 and 2F-140 top-level xacros were identical apart from the macro they include. The body now lives in `robotiq_2f_gripper.urdf.xacro`, keyed by a `model` arg (`2f_85` / `2f_140`); the two per-model files are 4-line wrappers, so existing `model:=` launch paths keep working.
- The `<hardware>` block of the two `ros2_control` xacros (plugin selection + driver params) was copied and had drifted: only the 2F-85 declared `use_dummy` and the Isaac `trigger_joint_command_threshold`. It now lives in one macro in `2f_common.ros2_control.xacro`, with a second macro for the state-only mimic joints the simulators need.
- Every launch argument added lately (`baudrate`, and the `sim_*` args in #47) had to be typed into four files; after this it is one top-level file plus one hardware macro. #47 will be rebased onto this.

## Behaviour
- 2F-85: expanded URDF byte-identical to `main` under mock, driver, sim_isaac and sim_gazebo (checked with `diff` on the `xacro` output).
- 2F-140: gains `use_dummy=false`, which is the driver's default when the param is absent, and the same Isaac threshold as the 2F-85. Nothing else changes.

## Test plan
- [x] `pytest test/test_ros2_control_urdf.py` passes (11 tests)
- [x] `xacro` output diffed against `main` for both models under all four hardware plugins
- [x] `ros2 launch robotiq_description robotiq_control.launch.py use_fake_hardware:=true` on Jazzy (in the `robotiq_ros2:jazzy` image with this branch's `urdf/`, `launch/`, `config/` mounted over the installed package): mock plugin loads, `joint_state_broadcaster`, `robotiq_gripper_controller` and `robotiq_activation_controller` all report `active`
- [x] same with `model:=.../robotiq_2f_140_gripper.urdf.xacro`: launches, but `robotiq_gripper_controller` fails to activate. Identical result on `main`, so pre-existing and unrelated to this PR: every `config/robotiq_controllers*.yaml` hardcodes `joint: robotiq_85_left_knuckle_joint`, and the 2F-140 drives `finger_joint`. Fixed in #49, which is stacked on this PR.

### DIFF
--- a/grippers/robotiq_description/urdf/2f_140.ros2_control.xacro
+++ b/grippers/robotiq_description/urdf/2f_140.ros2_control.xacro
@@ -1,6 +1,8 @@
 <?xml version="1.0"?>
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro">
 
+    <xacro:include filename="$(find robotiq_description)/urdf/2f_common.ros2_control.xacro" />
+
     <xacro:macro name="robotiq_gripper_ros2_control" params="
         name
         prefix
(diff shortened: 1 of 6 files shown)


jlongvalRobotiq:
Superseded by #50, same commits, opened as an in-repo branch so the stack (#50 → #51 → #52) shows only each PR's own commits.

PR #49, opened 2026-09-02
fix(robotiq_description): take the gripper joint from the launch so the 2F-140 activates

jlongvalRobotiq:
## Summary
- Both controller configs hardcoded the 2F-85's `robotiq_85_left_knuckle_joint` (as `joint`, `max_effort_interface`, `max_velocity_interface`), so a 2F-140 launch loaded `robotiq_gripper_controller` against a joint that does not exist and it never activated. Found while testing #48.
- The configs now name the joint through `$(var gripper_joint)`, resolved by `launch_ros` `ParameterFile(allow_substs=True)` for `ros2_control_node` and for each spawner's `--param-file`.
- New `gripper_joint` launch argument. Default: the 2F-85 joint, or `finger_joint` when the `model` path names a 2F-140, so the one-argument 2F-140 bringup the README describes actually works. Custom descriptions set it explicitly.
- README: the 2F-140 paragraph said to pass `description_file`; the argument is `model`. Fixed and documents `gripper_joint`.

## Behaviour change
The config files are no longer usable on their own: fed straight to `ros2_control_node` or a spawner without going through `robotiq_control.launch.py`, the joint stays the literal `$(var gripper_joint)` and the controller does not activate. Nothing in this repo does that (the tactile launch includes the control launch), but a downstream launch file pointing at our YAML would have to pass through the launch or set the joint itself.

Stacked on #48 (its two commits show here until it merges; GitHub cannot base a fork PR on another fork branch). #47 is rebased onto this.

## Test plan
- [x] `pytest test/` in `robotiq_description`: 30 passed, including a new check that the configs no longer hardcode the joint, that the launch defaults it from the model path, and that the launch description builds (the last one catches missing imports that only surface under `ros2 launch`)
- [x] `ros2 launch robotiq_description robotiq_control.launch.py use_fake_hardware:=true` in the `robotiq_ros2:jazzy` image (this branch's `urdf/`, `launch/`, `config/` mounted over the installed package): all three controllers `active`
- [x] same with `model:=.../robotiq_2f_140_gripper.urdf.xacro`: all three controllers `active` (was `inactive` for `robotiq_gripper_controller` on `main`); `ros2 param get /robotiq_gripper_controller joint` reads `finger_joint`
- [x] Humble: `ros:humble-ros-base` + `ros-humble-ros2-control` / `ros-humble-ros2-controllers`, `robotiq_controllers` and `robotiq_description` built from this branch. Both models with `use_fake_hardware:=true`: all three controllers `active` (`position_controllers/GripperActionController` from `gripper_controllers`); `joint` reads `robotiq_85_left_knuckle_joint` for the 2F-85 and `finger_joint` for the 2F-140

### DIFF
--- a/grippers/robotiq_description/urdf/robotiq_2f_gripper.urdf.xacro
+++ b/grippers/robotiq_description/urdf/robotiq_2f_gripper.urdf.xacro
@@ -1,0 +1,16 @@
+<?xml version="1.0"?>
+<robot xmlns:xacro="http://wiki.ros.org/xacro" name="robotiq_gripper">
+    <!-- parameters -->
+    <xacro:arg name="model" default="2f_85" />
+    <xacro:arg name="use_fake_hardware" default="true" />
+    <xacro:arg name="com_port" default="/dev/ttyUSB0" />
+    <xacro:arg name="baudrate" default="115200" />
+
+    <!-- Import macros -->
+    <xacro:include filename="$(find robotiq_description)/urdf/robotiq_$(arg model)_macro.urdf.xacro" />
+
+    <link name="world" />
+    <xacro:robotiq_gripper name="RobotiqGripperHardwareInterface" prefix="" parent="world" use_fake_hardware="$(arg use_fake_hardware)" com_port="$(arg com_port)" baudrate="$(arg baudrate)">
+        <origin xyz="0 0 0" rpy="0 0 0" />
+    </xacro:robotiq_gripper>
+</robot>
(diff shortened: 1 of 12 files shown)


grippers/robotiq_description/urdf/robotiq_2f_gripper.urdf.xacro
```diff
@@ -0,0 +1,16 @@
...
+    <xacro:arg name="baudrate" default="115200" />
+
+    <!-- Import macros -->
+    <xacro:include filename="$(find robotiq_description)/urdf/robotiq_$(arg model)_macro.urdf.xacro" />
+
+    <link name="world" />
+    <xacro:robotiq_gripper name="RobotiqGripperHardwareInterface" prefix="" parent="world" use_fake_hardware="$(arg use_fake_hardware)" com_port="$(arg com_port)" baudrate="$(arg baudrate)">
+        <origin xyz="0 0 0" rpy="0 0 0" />
+    </xacro:robotiq_gripper>
+</robot>
```
ebarnett3, line 16: nice job eliminating this duplication

jlongvalRobotiq:
Superseded by #51, same commits, opened as an in-repo branch so the stack (#50 → #51 → #52) shows only each PR's own commits.

PR #50, opened 2026-09-03
refactor(robotiq_description): share the top-level and ros2_control xacros between the 2F models

jlongvalRobotiq:
## Summary
- The 2F-85 and 2F-140 top-level xacros were identical apart from the macro they include. The body now lives in `robotiq_2f_gripper.urdf.xacro`, keyed by a `model` arg (`2f_85` / `2f_140`); the two per-model files are 4-line wrappers, so existing `model:=` launch paths keep working.
- The `<hardware>` block of the two `ros2_control` xacros (plugin selection + driver params) was copied and had drifted: only the 2F-85 declared `use_dummy` and the Isaac `trigger_joint_command_threshold`. It now lives in one macro in `2f_common.ros2_control.xacro`, with a second macro for the state-only mimic joints the simulators need.
- Every launch argument added lately (`baudrate`, and the `sim_*` args in #52) had to be typed into four files; after this it is one top-level file plus one hardware macro. #52 is stacked on #51, which is stacked on this.

## Behaviour
- 2F-85: expanded URDF byte-identical to `main` under mock, driver, sim_isaac and sim_gazebo (checked with `diff` on the `xacro` output).
- 2F-140: gains `use_dummy=false`, which is the driver's default when the param is absent, and the same Isaac threshold as the 2F-85. Nothing else changes.

## Test plan
- [x] `pytest test/test_ros2_control_urdf.py` passes (11 tests)
- [x] `xacro` output diffed against `main` for both models under all four hardware plugins
- [x] `ros2 launch robotiq_description robotiq_control.launch.py use_fake_hardware:=true` on Jazzy (in the `robotiq_ros2:jazzy` image with this branch's `urdf/`, `launch/`, `config/` mounted over the installed package): mock plugin loads, `joint_state_broadcaster`, `robotiq_gripper_controller` and `robotiq_activation_controller` all report `active`
- [x] same with `model:=.../robotiq_2f_140_gripper.urdf.xacro`: launches, but `robotiq_gripper_controller` fails to activate. Identical result on `main`, so pre-existing and unrelated to this PR: every `config/robotiq_controllers*.yaml` hardcodes `joint: robotiq_85_left_knuckle_joint`, and the 2F-140 drives `finger_joint`. Fixed in #51, which is stacked on this PR.

### DIFF
--- a/grippers/robotiq_description/urdf/robotiq_2f_140_gripper.urdf.xacro
+++ b/grippers/robotiq_description/urdf/robotiq_2f_140_gripper.urdf.xacro
@@ -1,15 +1,5 @@
 <?xml version="1.0"?>
 <robot xmlns:xacro="http://wiki.ros.org/xacro" name="robotiq_gripper">
-    <!-- parameters -->
-    <xacro:arg name="use_fake_hardware" default="true" />
-    <xacro:arg name="com_port" default="/dev/ttyUSB0" />
-    <xacro:arg name="baudrate" default="115200" />
-
-    <!-- Import macros -->
-    <xacro:include filename="$(find robotiq_description)/urdf/robotiq_2f_140_macro.urdf.xacro" />
-
-    <link name="world" />
-    <xacro:robotiq_gripper name="RobotiqGripperHardwareInterface" prefix="" parent="world" use_fake_hardware="$(arg use_fake_hardware)" com_port="$(arg com_port)" baudrate="$(arg baudrate)">
-        <origin xyz="0 0 0" rpy="0 0 0" />
-    </xacro:robotiq_gripper>
+    <xacro:arg name="model" default="2f_140" />
+    <xacro:include filename="$(find robotiq_description)/urdf/robotiq_2f_gripper.urdf.xacro" />
 </robot>
--- a/grippers/robotiq_description/urdf/robotiq_2f_140_gripper.urdf.xacro
+++ b/grippers/robotiq_description/urdf/robotiq_2f_140_gripper.urdf.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
 <robot xmlns:xacro="http://wiki.ros.org/xacro" name="robotiq_gripper">
-    <xacro:arg name="model" default="2f_140" />
+    <xacro:arg name="gripper_model" default="2f_140" />
     <xacro:include filename="$(find robotiq_description)/urdf/robotiq_2f_gripper.urdf.xacro" />
 </robot>
--- a/grippers/robotiq_description/urdf/robotiq_2f_gripper.urdf.xacro
+++ b/grippers/robotiq_description/urdf/robotiq_2f_gripper.urdf.xacro
@@ -1,0 +1,16 @@
+<?xml version="1.0"?>
+<robot xmlns:xacro="http://wiki.ros.org/xacro" name="robotiq_gripper">
+    <!-- parameters -->
+    <xacro:arg name="model" default="2f_85" />
+    <xacro:arg name="use_fake_hardware" default="true" />
+    <xacro:arg name="com_port" default="/dev/ttyUSB0" />
+    <xacro:arg name="baudrate" default="115200" />
+
+    <!-- Import macros -->
+    <xacro:include filename="$(find robotiq_description)/urdf/robotiq_$(arg model)_macro.urdf.xacro" />
+
+    <link name="world" />
+    <xacro:robotiq_gripper name="RobotiqGripperHardwareInterface" prefix="" parent="world" use_fake_hardware="$(arg use_fake_hardware)" com_port="$(arg com_port)" baudrate="$(arg baudrate)">
+        <origin xyz="0 0 0" rpy="0 0 0" />
+    </xacro:robotiq_gripper>
+</robot>
--- a/grippers/robotiq_description/urdf/robotiq_2f_gripper.urdf.xacro
+++ b/grippers/robotiq_description/urdf/robotiq_2f_gripper.urdf.xacro
@@ -1,13 +1,16 @@
 <?xml version="1.0"?>
 <robot xmlns:xacro="http://wiki.ros.org/xacro" name="robotiq_gripper">
     <!-- parameters -->
-    <xacro:arg name="model" default="2f_85" />
+    <!-- gripper_model selects the macro (2f_85 / 2f_140). xacro args are
+         first-declaration-wins, so each wrapper must declare it BEFORE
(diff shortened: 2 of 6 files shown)


grippers/robotiq_description/urdf/robotiq_2f_gripper.urdf.xacro
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
+<robot xmlns:xacro="http://wiki.ros.org/xacro" name="robotiq_gripper">
+    <!-- parameters -->
+    <xacro:arg name="model" default="2f_85" />
```
ebarnett3: the xacro `model` arg
   (`2f_85` / `2f_140`) shares its name with the launch `model` arg (a path,
   `D/launch/robotiq_control.launch.py:87`). `variant` for the xacro arg
   would avoid the confusion.

ebarnett3 (approved):

grippers/robotiq_description/urdf/robotiq_2f_gripper.urdf.xacro
mbegin-robotiq: Could we rename this arg to `gripper_model` (or `variant`)? As it stands `model` now has two incompatible meanings in the same package.

Both launch files already declare a `model` argument that is an **absolute path to a xacro file**, and substitute it as the file operand of the xacro command:

```python
DeclareLaunchArgument(
    name="model",
    default_value=default_model_path,   # .../urdf/robotiq_2f_85_gripper.urdf.xacro
    description="Absolute path to gripper URDF file",
)
```

So the command line a reader of this file will naturally reach for fails:

```console
$ ros2 launch robotiq_description view_gripper.launch.py model:=2f_140
```

That overrides the *launch* arg, i.e. the path operand, so the launch ends up running `xacro 2f_140`:

```
FileNotFoundError: [Errno 2] No such file or directory: './2f_140'
```

Picking the 140 through launch still requires the full path. And since neither launch file forwards `model:=` into the xacro command (only `use_fake_hardware`, `com_port`, `baudrate`), the xacro-level arg is unreachable from the launch layer entirely — it is settable only by invoking `xacro` directly, which does work: `xacro robotiq_2f_gripper.urdf.xacro model:=2f_140` resolves the include to `robotiq_2f_140_macro.urdf.xacro` as intended.

The rename is mechanical — three sites: this `xacro:arg`, the `$(arg model)` on line 10, and the matching line in each of the two wrappers. Nothing outside these three xacros reads it, and it leaves `model` free to keep meaning "path to a xacro file" for every existing caller.

grippers/robotiq_description/urdf/robotiq_2f_140_gripper.urdf.xacro
```diff
@@ -1,15 +1,5 @@
...
-    <xacro:include filename="$(find robotiq_description)/urdf/robotiq_2f_140_macro.urdf.xacro" />
-
-    <link name="world" />
-    <xacro:robotiq_gripper name="RobotiqGripperHardwareInterface" prefix="" parent="world" use_fake_hardware="$(arg use_fake_hardware)" com_port="$(arg com_port)" baudrate="$(arg baudrate)">
-        <origin xyz="0 0 0" rpy="0 0 0" />
-    </xacro:robotiq_gripper>
-</robot>
+<?xml version="1.0"?>
+<robot xmlns:xacro="http://wiki.ros.org/xacro" name="robotiq_gripper">
+    <xacro:arg name="model" default="2f_140" />
```
mbegin-robotiq: Minor / your call, no change needed for correctness.

This declaration before the include is what makes the indirection work — xacro args are first-declaration-wins, so the wrapper's value wins and the `default="2f_85"` in `robotiq_2f_gripper.urdf.xacro` is inert in normal use.

The flip side is that the contract is implicit: a future wrapper (a 2F-105, say) that omits this line would silently build an 85 rather than failing. Two ways to close that off, whichever you prefer:

- drop the `default` in `robotiq_2f_gripper.urdf.xacro` so an omission errors out loudly, or
- keep it and add a one-liner there noting that each wrapper must declare the arg ahead of the include.

Happy to leave it as is if you would rather not touch it.

mbegin-robotiq (approved):
Behaviour-preserving across all four hardware modes for both models as far as I can tell — I expanded both top-level xacros at base and at this tip in each mode and diffed the rendered URDFs, and `test_ros2_control_urdf.py` is green (11 passed). The `$(arg model)`-in-include-path trick resolves correctly (xacro args are first-declaration-wins, so the 140 wrapper really does pull in the 140 macro and meshes), and no macro-redefinition warnings appear. The CRLF cleanup in `robotiq_2f_140_gripper.urdf.xacro` is a nice bonus.

Approving. Two comments inline — the first is the one I would like addressed before merge, the second is your call.

One non-blocking nit: the new shared file's header comment says the `<hardware>` block "is the same for every 2F model", which is not quite right now — the rendered 140 picks up `trigger_joint_command_threshold` and `use_dummy` that it did not emit before. Both are harmless (0.02 rad is essentially the same relative deadband given the 140's `0 → 0.7` `finger_joint` range as it is for the 85's `0 → 0.8`, and `use_dummy=false` parses identically to the param being absent), so this is a wording fix, not a code change.

grippers/robotiq_description/urdf/robotiq_2f_gripper.urdf.xacro
jlongvalRobotiq: Renamed to `gripper_model` in a194280 (the three xacro sites; nothing else read it). `model` keeps meaning "path to a xacro file" at the launch layer. Rendered 2F-85 and 2F-140 URDFs are byte-identical to before.

grippers/robotiq_description/urdf/robotiq_2f_140_gripper.urdf.xacro
jlongvalRobotiq: Tried the first option: xacro refuses a `xacro:arg` without a `default` attribute ("missing attribute 'default'"), so a loud failure is not available. Went with the second: default kept, plus a note above the arg that each wrapper must declare it ahead of the include.

jlongvalRobotiq:
Header reworded in f10b821: it now just says the block is shared and where the per-model parts live, without claiming the output is identical per model.

grippers/robotiq_description/urdf/robotiq_2f_gripper.urdf.xacro
jlongvalRobotiq: Renamed to `gripper_model` in a194280, see the thread below for details.

PR #51, opened 2026-09-03
fix(robotiq_description): take the gripper joint from the launch so the 2F-140 activates

jlongvalRobotiq:
## Summary
- Both controller configs hardcoded the 2F-85's `robotiq_85_left_knuckle_joint` (as `joint`, `max_effort_interface`, `max_velocity_interface`), so a 2F-140 launch loaded `robotiq_gripper_controller` against a joint that does not exist and it never activated. Found while testing #50.
- The configs now name the joint through `$(var gripper_joint)`, resolved by `launch_ros` `ParameterFile(allow_substs=True)` for `ros2_control_node` and for each spawner's `--param-file`.
- New `gripper_model` launch argument (`2f_85` / `2f_140`) selects the gripper macro in the xacro and the joint the controller drives, so a 2F-140 bringup is `gripper_model:=2f_140`. `gripper_joint` still exists for custom descriptions. Replaces an earlier version that sniffed the `model` path for `2f_140`.
- README: the 2F-140 paragraph said to pass `description_file`; the argument is `model`. Fixed and documents `gripper_joint`.

## Behaviour change
The config files are no longer usable on their own: fed straight to `ros2_control_node` or a spawner without going through `robotiq_control.launch.py`, the joint stays the literal `$(var gripper_joint)` and the controller does not activate. Nothing in this repo does that (the tactile launch includes the control launch), but a downstream launch file pointing at our YAML would have to pass through the launch or set the joint itself.

#50 is merged; this PR now sits on `main`. #52 is stacked on this.

## Test plan
- [x] `pytest test/` in `robotiq_description`: 30 passed, including a new check that the configs no longer hardcode the joint, that the launch defaults it from `gripper_model`, forwards `gripper_model` to xacro and restricts it to the known models, and that the launch description builds (the last one catches missing imports that only surface under `ros2 launch`)
- [x] `ros2 launch robotiq_description robotiq_control.launch.py use_fake_hardware:=true` in the `robotiq_ros2:jazzy` image (this branch's `urdf/`, `launch/`, `config/` mounted over the installed package): all three controllers `active`
- [x] same with `gripper_model:=2f_140`: all three controllers `active` (was `inactive` for `robotiq_gripper_controller` on `main`); `ros2 param get /robotiq_gripper_controller joint` reads `finger_joint`
- [x] Humble: `ros:humble-ros-base` + `ros-humble-ros2-control` / `ros-humble-ros2-controllers`, `robotiq_controllers` and `robotiq_description` built from this branch. Both models with `use_fake_hardware:=true`: all three controllers `active` (`position_controllers/GripperActionController` from `gripper_controllers`); `joint` reads `robotiq_85_left_knuckle_joint` for the 2F-85 and `finger_joint` for the 2F-140

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,8 +133,11 @@ In the combined launch the pad frames are TF-mounted on the gripper fingertip li
 ROS 2 `ros2_control` driver for Robotiq 2F adaptive grippers, under [`grippers/`](grippers/).
 
 Descriptions ship for the **2F-85** and the **2F-140**; `robotiq_control.launch.py` defaults to the
-2F-85, so pass `description_file` for a 2F-140. Hardware validation to date is on a 2F-85 — the
-2F-140 description ships untested against hardware.
+2F-85, so pass `model:=$(ros2 pkg prefix robotiq_description)/share/robotiq_description/urdf/robotiq_2f_140_gripper.urdf.xacro`
+for a 2F-140. The controller configs take the driven joint from the `gripper_joint` launch argument,
+which defaults to the 2F-85's `robotiq_85_left_knuckle_joint`, or to `finger_joint` when the model
+path names a 2F-140; set it explicitly for a custom description. Hardware validation to date is on a
+2F-85 — the 2F-140 description ships untested against hardware.
 
 The driver itself is model-agnostic: it needs a serial link and the `gripper_closed_position` of
 whatever is attached. A **Hand-E** therefore works once you supply a URDF for it, but no Hand-E
--- a/README.md
+++ b/README.md
@@ -133,11 +133,12 @@ In the combined launch the pad frames are TF-mounted on the gripper fingertip li
 ROS 2 `ros2_control` driver for Robotiq 2F adaptive grippers, under [`grippers/`](grippers/).
 
 Descriptions ship for the **2F-85** and the **2F-140**; `robotiq_control.launch.py` defaults to the
-2F-85, so pass `model:=$(ros2 pkg prefix robotiq_description)/share/robotiq_description/urdf/robotiq_2f_140_gripper.urdf.xacro`
-for a 2F-140. The controller configs take the driven joint from the `gripper_joint` launch argument,
-which defaults to the 2F-85's `robotiq_85_left_knuckle_joint`, or to `finger_joint` when the model
-path names a 2F-140; set it explicitly for a custom description. Hardware validation to date is on a
-2F-85 — the 2F-140 description ships untested against hardware.
+2F-85, so pass `gripper_model:=2f_140` for a 2F-140. That argument selects the gripper macro in the
+xacro and the joint the controller drives (`robotiq_85_left_knuckle_joint` or `finger_joint`); a
+custom description keeps `model:=<path>` and sets `gripper_joint` explicitly. The controller configs
+in `config/` write that joint as `$(var gripper_joint)`, which only the launch resolves — loaded
+directly into your own `ros2_control_node` they are not valid as they stand. Hardware validation to
+date is on a 2F-85 — the 2F-140 description ships untested against hardware.
 
 The driver itself is model-agnostic: it needs a serial link and the `gripper_closed_position` of
 whatever is attached. A **Hand-E** therefore works once you supply a URDF for it, but no Hand-E
--- a/grippers/robotiq_description/config/robotiq_controllers.humble.yaml
+++ b/grippers/robotiq_description/config/robotiq_controllers.humble.yaml
@@ -27,7 +27,7 @@ controller_manager:
 
 robotiq_gripper_controller:
   ros__parameters:
-    joint: robotiq_85_left_knuckle_joint
+    joint: $(var gripper_joint)
 
     max_effort: 50.0   # TODO tune
 
--- a/grippers/robotiq_description/launch/robotiq_control.launch.py
+++ b/grippers/robotiq_description/launch/robotiq_control.launch.py
@@ -27,14 +27,17 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 import launch
+from launch.substitution import Substitution
 from launch.substitutions import (
     Command,
     FindExecutable,
     LaunchConfiguration,
     PathJoinSubstitution,
+    PythonExpression,
 )
 from launch.conditions import IfCondition
 import launch_ros
+from launch_ros.parameter_descriptions import ParameterFile
 import os
 
 # Humble has no parallel_gripper_controller package, so it needs its own
@@ -46,6 +49,25 @@ HUMBLE_CONTROLLERS_FILE = "robotiq_controllers.humble.yaml"
 def controllers_file_for_distro(distro):
     """Return the controller config file name to load on ROS distro `distro`."""
     return HUMBLE_CONTROLLERS_FILE if distro == "humble" else JAZZY_CONTROLLERS_FILE
+
+
+class ParameterFilePath(Substitution):
+    def __init__(self, parameter_file):
+        super().__init__()
+        self.parameter_file = parameter_file
+
+    def perform(self, context):
+        return str(self.parameter_file.evaluate(context))
+
+
+def default_gripper_joint():
+    return PythonExpression(
+        [
+            "'finger_joint' if '2f_140' in '",
+            LaunchConfiguration("model"),
+            "' else 'robotiq_85_left_knuckle_joint'",
+        ]
+    )
 
 
 def generate_launch_description():
@@ -77,6 +99,13 @@ def generate_launch_description():
     args.append(
         launch.actions.DeclareLaunchArgument(
             name="launch_rviz", default_value="false", description="Launch RViz?"
+        )
+    )
+    args.append(
+        launch.actions.DeclareLaunchArgument(
+            name="gripper_joint",
+            default_value=default_gripper_joint(),
+            description="Joint the gripper controller drives; defaults from the model",
         )
     )
     args.append(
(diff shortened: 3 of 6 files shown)


grippers/robotiq_description/config/robotiq_controllers.humble.yaml
```diff
@@ -27,7 +27,7 @@ controller_manager:
 
 robotiq_gripper_controller:
   ros__parameters:
-    joint: robotiq_85_left_knuckle_joint
+    joint: $(var gripper_joint)
```
ebarnett3, line 30: **Config files stop working outside the launch (judgment call).**
   `D/config/robotiq_controllers.yaml:17,20,24` and
   `D/config/robotiq_controllers.humble.yaml:30`. The file keeps PickNik's
   filename, and a downstream launch that loads it straight into its own
   `ros2_control_node` now gets a controller that fails to activate on a
   literal `$(var gripper_joint)`. The author flagged it under "Behaviour
   change".

   Alternative that keeps the files literal: leave the 2F-85 joint in the
   YAML and have the launch rewrite it. `ParameterFilePath`
   (`D/launch/robotiq_control.launch.py:54`) already exists; make it load the
   YAML, set the joint keys from `gripper_joint`, dump to a
   `NamedTemporaryFile` and return the path. Same size as the current class,
   standalone use keeps working for the 2F-85.

   If the break is accepted: say so in the README paragraph that documents
   `gripper_joint` (`README.md:137`).

grippers/robotiq_description/test/test_ros2_control_urdf.py
```diff
@@ -157,10 +157,12 @@ def test_gripper_controller_config_interfaces_exist_under_mock():
     params = yaml.safe_load(CONFIG.read_text())["robotiq_gripper_controller"][
         "ros__parameters"
     ]
-    claimed = {params["max_effort_interface"], params["max_velocity_interface"]}
-
     ros2_control = expand("robotiq_2f_85_gripper.urdf.xacro", use_fake_hardware=True)
```
ebarnett3: **Coverage misses the case the PR fixes.**
   `D/test/test_ros2_control_urdf.py:153-160` expands only the 2F-85.
   Parametrize over `MODELS` so the 2F-140 interface names are checked
   against its mock URDF.

grippers/robotiq_description/test/test_controller_config.py
ebarnett3: **`test_joint_matches_exported_interfaces` is now tautological.**
   `D/test/test_controller_config.py:121`; `gripper_controller_params()`
   (`:88`) substitutes `JOINT` for the placeholder and the test asserts the
   result equals `JOINT`. Drop it or parametrize over both joints. (#52
   extends it to the sim configs, so fix it here.)

README.md
```diff
@@ -133,8 +133,11 @@ In the combined launch the pad frames are TF-mounted on the gripper fingertip li
 ROS 2 `ros2_control` driver for Robotiq 2F adaptive grippers, under [`grippers/`](grippers/).
 
 Descriptions ship for the **2F-85** and the **2F-140**; `robotiq_control.launch.py` defaults to the
-2F-85, so pass `description_file` for a 2F-140. Hardware validation to date is on a 2F-85 — the
-2F-140 description ships untested against hardware.
+2F-85, so pass `model:=$(ros2 pkg prefix robotiq_description)/share/robotiq_description/urdf/robotiq_2f_140_gripper.urdf.xacro`
+for a 2F-140. The controller configs take the driven joint from the `gripper_joint` launch argument,
+which defaults to the 2F-85's `robotiq_85_left_knuckle_joint`, or to `finger_joint` when the model
```
ebarnett3: `robotiq_85_left_knuckle_joint` for 2F-85 and `finger_joint` for 2F-140 is inconsistent nomenclature. But this was not introduced here so we can address it in future work.

ebarnett3 (approved):

grippers/robotiq_description/launch/robotiq_control.launch.py
```diff
@@ -48,6 +51,25 @@ def controllers_file_for_distro(distro):
...
+        self.parameter_file = parameter_file
+
+    def perform(self, context):
+        return str(self.parameter_file.evaluate(context))
+
+
+def default_gripper_joint():
+    return PythonExpression(
+        [
+            "'finger_joint' if '2f_140' in '",
```
mbegin-robotiq: **The variant match is looser than it looks — and I think it belongs on #50's rename.**

`'2f_140' in '<model>'` tests the *full path*, not the filename, and does it case-sensitively. Two ways it silently picks the wrong joint:

- a 2F-85 model living under a `2f_140_cell/` (or `.../2f_140/`) directory gets `finger_joint`
- `robotiq_2F_140_gripper.urdf.xacro` — the capitalisation the product name actually uses — misses the match and gets `robotiq_85_left_knuckle_joint`

Either way the controller fails to activate on a joint the hardware does not export, which is precisely the failure this PR exists to fix, just relocated.

Tightening the match (basename, case-insensitive) would close both, but I would rather not harden a substring sniff that shouldn't be here. #50 is carrying a thread about `model` being overloaded — an absolute path at the launch layer, a variant (`2f_85` / `2f_140`) at the xacro layer — with a rename to `variant` / `gripper_model` proposed. This function is a third consumer, recovering the variant by grepping the path because no variant argument is reachable from launch.

If that rename lands, `default_gripper_joint()` reads the variant directly and the whole class of bug goes away. Worth sequencing behind #50 rather than patching the regex here.

grippers/robotiq_description/launch/robotiq_control.launch.py
```diff
@@ -48,6 +51,25 @@ def controllers_file_for_distro(distro):
...
+    def __init__(self, parameter_file):
+        super().__init__()
+        self.parameter_file = parameter_file
+
+    def perform(self, context):
+        return str(self.parameter_file.evaluate(context))
+
+
+def default_gripper_joint():
+    return PythonExpression(
```
mbegin-robotiq: **An apostrophe in the model path aborts the launch with a bare `SyntaxError`.**

The path is interpolated unescaped between single quotes, and `PythonExpression` `eval`s the concatenated string. So `model:=/home/o'brien/ws/.../robotiq_2f_85_gripper.urdf.xacro` renders as:

```python
'finger_joint' if '2f_140' in '/home/o'brien/ws/...' else 'robotiq_85_left_knuckle_joint'
```

which fails to parse. The traceback points at the expression, not at the path, so it reads as a bug in the launch file rather than "your home directory has an apostrophe in it".

Rare, but it costs nothing to avoid — and it disappears entirely if this ends up keying off a variant argument rather than the path (see the note below on #50).

mbegin-robotiq (approved):
Approving — the mechanism is right and I verified it end to end against Jazzy rather than taking it on faith.

`ParameterFile(..., allow_substs=True)` does rewrite `$(var gripper_joint)` in both YAML configs, including inside the quoted `max_*_interface` strings; nothing else in either file trips the substitution parser (the `$ROS_DISTRO` in the comments survives verbatim); `ParameterFile.evaluate` caches, so the control node and all three spawners share one temp file rather than racing on separate ones; and the spawner/controller_manager path imposes no `.yaml` suffix requirement on that temp file. On the hardware side, the driver exports its command interfaces from `info_.joints[0].name` rather than a hardcoded 2F-85 name, so `finger_joint` is correct on a real 2F-140 and not just under mock. Argument declaration order is fine — `model` is declared before `gripper_joint`, so the default resolves. 19 tests green in `test_controller_config.py`.

Two non-blocking notes left inline on `default_gripper_joint()`. The substring match against the full model path is the one I care about: it is loose enough to pick the wrong joint for a 2F-85 under a `2f_140_cell/` directory, or for a `robotiq_2F_140_*.xacro`. I would not fix it by tightening the match — #50 is already discussing the `model` overload, and if that rename lands this function reads the variant directly and the bug class disappears. Happy for it to land as is and be cleaned up in that sequence.

Not merging — leaving that to you, and #50 is underneath this one anyway.

grippers/robotiq_description/config/robotiq_controllers.humble.yaml
jlongvalRobotiq: Accepted the break and documented it in 25683f0: the README now says the configs write the joint as `$(var gripper_joint)`, which only the launch resolves, so they are not valid loaded directly into another `ros2_control_node`. Kept `ParameterFile(allow_substs=True)` over a temp-file rewrite since mbegin verified that path end to end below.

grippers/robotiq_description/test/test_ros2_control_urdf.py
jlongvalRobotiq: Parametrized over `MODELS` in b696511; the 2F-140 mock URDF is now checked against the config too.

grippers/robotiq_description/test/test_controller_config.py
jlongvalRobotiq: Dropped in b696511. The two 2F-85-only interface tests became one test over both joints that substitutes each into the config and checks the claimed names against what the driver exports for that joint. #52 will get the same treatment when it is rebased.

README.md
jlongvalRobotiq: Agreed, left as is for now.

grippers/robotiq_description/launch/robotiq_control.launch.py
jlongvalRobotiq: Went with your sequencing. 6d5554b adds a `gripper_model` launch argument (`2f_85` / `2f_140`, enforced with `choices`), forwards it to xacro so both layers agree, and looks the default joint up in a dict. No path sniffing left. A 2F-140 is now `gripper_model:=2f_140` with `model` at its default; the `model` path is only needed for custom descriptions, which set `gripper_joint` themselves.

grippers/robotiq_description/launch/robotiq_control.launch.py
jlongvalRobotiq: Gone with 6d5554b: the default joint is a plain `Substitution` doing a dict lookup, so nothing user-supplied is spliced into an `eval` anymore.

Issue #29, opened 2026-08-14
robotiq_driver: velocity state is always 0.0, so every gripper_cmd goal reports stalled / reached_goal false


ebarnett3:
The `velocity` state interface is never computed — it is registered, set to `0` once, and stays there. `parallel_gripper_action_controller` derives stall detection from that value, so **every** goal trips stall detection and reports `stalled: true` / `reached_goal: false`, even when the gripper reaches the commanded position exactly.

Pre-existing, not a regression: PickNik's pre-SDK driver had the same three references to `gripper_velocity_` (NaN at construction, the state-interface registration, `= 0` in `on_activate`) and never updated it in `read()`. The SDK migration preserved the behaviour and made it explicit — `read()` now assigns `0.0` each cycle, since the gripper's status block carries position and motor current only.

## Measured on a 2F-85 (Jazzy, v1.1.0)

| goal | landed at | result |
|---|---|---|
| `position: [0.7]` | 0.69859 rad | `stalled: true`, `reached_goal: false` |
| `position: [0.6]` | 0.59729 rad | `stalled: true`, `reached_goal: false` |
| `position: [0.0]` | — | `stalled: true`, `reached_goal: false` |

Both landings are well inside `goal_tolerance` (0.02). Sampling `/joint_states` during an active move yields one distinct velocity value: `0.0`.

Controller defaults involved: `stall_velocity_threshold = 0.001`, `stall_timeout = 0.05`, `allow_stalling = true`. Because stalling is allowed, the action still finishes `SUCCEEDED` — so a client that checks the action status sees success, while a client that checks `result.reached_goal` sees false on every successful grasp.

Verified on Jazzy. Humble's `position_controllers/GripperActionController` uses the same velocity-based stall detection, so it is expected to behave the same way there; not yet measured.

## The obvious fix is the wrong one

Differentiating position in `read()` does not work as-is. `gPO` is 8 bits over roughly 0.8 rad, so one count is ~0.003 rad; at the default 100 Hz exchange most cycles show no change at all and the occasional count-flip produces a large spike. A raw derivative would therefore alternate between 0 and a spike, making stall detection *flap* rather than behave — arguably worse than the current consistent wrongness.

A real fix needs some combination of:

- filtering the derivative (and accepting the lag that adds),
- re-tuning `stall_velocity_threshold` / `stall_timeout` against the resulting signal on hardware,
[...]

PR #52, opened 2026-09-03
feat(robotiq_description): let the sim_isaac / sim_gazebo plugins activate from the launch

jlongvalRobotiq:
## Problem

The gripper xacros have let you pick `sim_isaac` (topic_based_ros2_control) as the hardware plugin for years, but the controller config and the launch file were written for the real driver only. Pick the sim plugin and `robotiq_gripper_controller` never activates: the config claims `set_gripper_max_effort` / `set_gripper_max_velocity`, which only the driver exports, and the launch spawns `robotiq_activation_controller`, whose GPIO the sim URDF does not declare. The launch also never exposed the sim flag at all.

## Change

One commit per layer, oldest first:

1. **xacro** — the top-level 2F-85 / 2F-140 xacros accept and forward `sim_isaac`, `isaac_joint_commands`, `isaac_joint_states` (the macro already had them).
2. **config** — `robotiq_controllers.sim.yaml` (+ Humble variant): same controllers as the hardware config minus the activation controller and the two driver-only command interfaces.
3. **launch** — new arguments, defaulting to today's behaviour; `sim_isaac` selects the sim config and skips the activation spawner.
4. **README** — the sim launch line and a note that `TopicBasedSystem` matches joints by name.
5. **review follow-ups** — `sim_gazebo` dropped from the launch surface (it could not work as wired; the macro-level parameter predates this PR and stays), `use_fake_hardware` + `sim_isaac` together now fails the launch by name, a stall in sim aborts the goal instead of succeeding, and the README lists both models' joint names.

Hardware and mock bringup are unchanged: same files, same controllers, same action surface. The sim plugin does not become a dependency.

## Verification

- 24 → 57 tests in `robotiq_description`; green at every commit. New tests pin what the sim plugin exports, that the sim config claims only that, the launch's own spawner wiring per distro, and the flag-conflict check.
- Jazzy, this tree: `use_fake_hardware:=true` with `gripper_model:=2f_85` and `:=2f_140` activates all three controllers with the right joint; `use_fake_hardware:=true sim_isaac:=true` fails at launch with the conflict message.
- Jazzy: `sim_isaac:=true` exercised live against an Isaac Sim 5.1 twin of a 2F-85 (open, close, close-on-object through `gripper_cmd`) before the stall change. With `allow_stalling: false` a close-on-object is expected to come back as an aborted goal with `stalled: true`, per the controller source; that path is next exercised by the MCP client work that consumes it.
- black, flake8, codespell clean.

## Stack
Stacked on #51: this PR shows only its own 9 commits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,9 +257,24 @@ ros2 launch robotiq_description robotiq_control.launch.py                    # r
 ros2 launch robotiq_description robotiq_control.launch.py use_fake_hardware:=true   # ros2_control mock
 ros2 launch robotiq_description robotiq_control.launch.py launch_rviz:=true         # + RViz visualization
 ros2 launch robotiq_description robotiq_control.launch.py baudrate:=<rate>
+ros2 launch robotiq_description robotiq_control.launch.py sim_gazebo:=true        # gz_ros2_control
+ros2 launch robotiq_description robotiq_control.launch.py sim_isaac:=true \
+  isaac_joint_commands:=/isaac_joint_commands isaac_joint_states:=/isaac_joint_states  # topic_based_ros2_control
 ```
 
 This activates `joint_state_broadcaster`, `robotiq_gripper_controller`, and `robotiq_activation_controller`.
+
+`sim_gazebo` and `sim_isaac` swap the hardware plugin for `gz_ros2_control/GazeboSimSystem` or
+`topic_based_ros2_control/TopicBasedSystem` (any simulator that exchanges `sensor_msgs/JointState`
+on two topics — Isaac Sim is the usual one, hence the argument names). Both plugins export only the
+standard joint interfaces, so the launch loads `config/robotiq_controllers.sim.yaml` for them
+(per-goal `effort`/`velocity` are accepted and ignored; stall detection is tuned for a loop that
+runs over a topic) and skips `robotiq_activation_controller`, whose `reactivate_gripper` GPIO only
+the driver and the mock declare. `TopicBasedSystem` matches joints to the simulator's `JointState`
+**by name**, so the simulator must publish this description's joint names (`robotiq_85_left_knuckle_joint`
+and the five mimic joints, with your `prefix`), or nothing moves. Neither plugin is a dependency of
+this package — install `ros-$ROS_DISTRO-gz-ros2-control` or build
+[topic_based_ros2_control](https://github.com/PickNikRobotics/topic_based_ros2_control) yourself.
 
 ### Commanding the gripper
 
--- a/README.md
+++ b/README.md
@@ -257,24 +257,25 @@ ros2 launch robotiq_description robotiq_control.launch.py                    # r
 ros2 launch robotiq_description robotiq_control.launch.py use_fake_hardware:=true   # ros2_control mock
 ros2 launch robotiq_description robotiq_control.launch.py launch_rviz:=true         # + RViz visualization
 ros2 launch robotiq_description robotiq_control.launch.py baudrate:=<rate>
-ros2 launch robotiq_description robotiq_control.launch.py sim_gazebo:=true        # gz_ros2_control
 ros2 launch robotiq_description robotiq_control.launch.py sim_isaac:=true \
   isaac_joint_commands:=/isaac_joint_commands isaac_joint_states:=/isaac_joint_states  # topic_based_ros2_control
 ```
 
 This activates `joint_state_broadcaster`, `robotiq_gripper_controller`, and `robotiq_activation_controller`.
 
-`sim_gazebo` and `sim_isaac` swap the hardware plugin for `gz_ros2_control/GazeboSimSystem` or
-`topic_based_ros2_control/TopicBasedSystem` (any simulator that exchanges `sensor_msgs/JointState`
-on two topics — Isaac Sim is the usual one, hence the argument names). Both plugins export only the
-standard joint interfaces, so the launch loads `config/robotiq_controllers.sim.yaml` for them
-(per-goal `effort`/`velocity` are accepted and ignored; stall detection is tuned for a loop that
-runs over a topic) and skips `robotiq_activation_controller`, whose `reactivate_gripper` GPIO only
-the driver and the mock declare. `TopicBasedSystem` matches joints to the simulator's `JointState`
-**by name**, so the simulator must publish this description's joint names (`robotiq_85_left_knuckle_joint`
-and the five mimic joints, with your `prefix`), or nothing moves. Neither plugin is a dependency of
-this package — install `ros-$ROS_DISTRO-gz-ros2-control` or build
+`sim_isaac` swaps the hardware plugin for `topic_based_ros2_control/TopicBasedSystem` (any
+simulator that exchanges `sensor_msgs/JointState` on two topics — Isaac Sim is the usual one, hence
+the argument names). That plugin exports only the standard joint interfaces, so the launch loads
+`config/robotiq_controllers.sim.yaml` for it (per-goal `effort`/`velocity` are accepted and ignored;
+stall detection is tuned for a loop that runs over a topic) and skips `robotiq_activation_controller`,
+whose `reactivate_gripper` GPIO only the driver and the mock declare. `TopicBasedSystem` matches
+joints to the simulator's `JointState` **by name**, so the simulator must publish this description's
+joint names (`robotiq_85_left_knuckle_joint` and the five mimic joints, with your `prefix`), or
+nothing moves. The plugin is not a dependency of this package — build
 [topic_based_ros2_control](https://github.com/PickNikRobotics/topic_based_ros2_control) yourself.
+The `ros2_control` xacros also carry a `sim_gazebo` parameter (`gz_ros2_control/GazeboSimSystem`)
+for cells that embed the gripper macro in their own Gazebo description; this launch does not wire
+it, since Gazebo hosts its own `controller_manager`.
 
 ### Commanding the gripper
 
--- a/README.md
+++ b/README.md
@@ -267,7 +267,8 @@ This activates `joint_state_broadcaster`, `robotiq_gripper_controller`, and `rob
 simulator that exchanges `sensor_msgs/JointState` on two topics — Isaac Sim is the usual one, hence
 the argument names). That plugin exports only the standard joint interfaces, so the launch loads
 `config/robotiq_controllers.sim.yaml` for it (per-goal `effort`/`velocity` are accepted and ignored;
-stall detection is tuned for a loop that runs over a topic) and skips `robotiq_activation_controller`,
+a stall aborts the goal rather than succeeding, since a simulator that publishes no joint velocities
+would otherwise report every failed grasp as success) and skips `robotiq_activation_controller`,
 whose `reactivate_gripper` GPIO only the driver and the mock declare. `TopicBasedSystem` matches
 joints to the simulator's `JointState` **by name**, so the simulator must publish this description's
 joint names (`robotiq_85_left_knuckle_joint` and the five mimic joints, with your `prefix`), or
--- a/README.md
+++ b/README.md
@@ -271,8 +271,14 @@ a stall aborts the goal rather than succeeding, since a simulator that publishes
 would otherwise report every failed grasp as success) and skips `robotiq_activation_controller`,
 whose `reactivate_gripper` GPIO only the driver and the mock declare. `TopicBasedSystem` matches
 joints to the simulator's `JointState` **by name**, so the simulator must publish this description's
-joint names (`robotiq_85_left_knuckle_joint` and the five mimic joints, with your `prefix`), or
-nothing moves. The plugin is not a dependency of this package — build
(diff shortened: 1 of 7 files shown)


mbegin-robotiq (commented):
Code review of the sim-plugin activation work. Four findings inline: the `sim_gazebo` path can't activate controllers as wired, the three hardware-selection flags aren't mutually exclusive, `allow_stalling` makes failed grasps report success on simulators that don't publish velocities, and the README joint-name note is 2F-85-only.

Checked and clean: the nested `PythonExpression` config selection (`evaluate_condition_expression` strips/lowercases the `"True"/"False"`), the new `condition=` plumbing on the spawner, and installation of the new YAMLs via the existing `install(DIRECTORY config ...)` rule. Tests pass locally (62) once `robotiq_description` is on the ament index — the stale copy in `/opt/ros/jazzy/share` otherwise breaks every xacro expansion.

grippers/robotiq_description/launch/robotiq_control.launch.py
```diff
@@ -129,6 +138,45 @@ def generate_launch_description():
...
+            default_value="false",
+            description="Drive a simulator over topic_based_ros2_control instead of a real gripper",
+        )
+    )
+    args.append(
+        launch.actions.DeclareLaunchArgument(
+            name="sim_gazebo",
+            default_value="false",
+            description="Drive Gazebo over gz_ros2_control instead of a real gripper",
+        )
```
mbegin-robotiq: `sim_gazebo:=true` is documented but can't work as wired. Two independent reasons:

1. The launch unconditionally starts a standalone `ros2_control_node`, which loads `gz_ros2_control/GazeboSimSystem` outside Gazebo. `initSim` is never called, so no interfaces are exported and the controllers never activate.
2. No xacro declares the `gz_ros2_control-system` Gazebo plugin, so Gazebo doesn't host a `controller_manager` either — there's no process on the other side that *could* activate them.

Only the `sim_isaac` path appears to have been verified live. Either wire the Gazebo path properly (declare the system plugin in the xacro and suppress the standalone `ros2_control_node` when `sim_gazebo` is set) or drop the argument and the README section until it's implemented.

grippers/robotiq_description/launch/robotiq_control.launch.py
```diff
@@ -129,6 +138,45 @@ def generate_launch_description():
             description="Use ros2_control mock (fake) hardware instead of a real gripper",
         )
     )
+    args.append(
```
mbegin-robotiq, line 220: `use_fake_hardware`, `sim_isaac` and `sim_gazebo` aren't mutually exclusive, and setting more than one silently produces a broken description. Confirmed by expanding the xacros by hand:

- two flags set → `<hardware>` gets two `<plugin>` elements;
- fake + isaac → the joint still declares `set_gripper_max_*` while the launch has switched to the sim config that no longer claims those interfaces.

The tests can't catch this today: `plugin_of` only reads the first `<plugin>`, and no test sets two flags. Worth an explicit check that at most one is true (raising at launch), plus a test for the combination.

grippers/robotiq_description/config/robotiq_controllers.sim.yaml
```diff
@@ -0,0 +1,37 @@
...
+      type: parallel_gripper_action_controller/GripperActionController
+
+robotiq_gripper_controller:
+  ros__parameters:
+    joint: $(var gripper_joint)
+
+    state_interfaces: ["position", "velocity"]
+    allow_stalling: true
+    stall_timeout: 0.5
+    stall_velocity_threshold: 0.05
```
mbegin-robotiq: `allow_stalling: true` with these thresholds makes failed grasps report success. A simulator that publishes no velocities in its `JointState` leaves the velocity state at 0, which is below `stall_velocity_threshold: 0.05` from the start — so any out-of-tolerance goal is reported **succeeded** (stalled) after `stall_timeout: 0.5`. Clients can't distinguish "gripped the object" from "nothing moved at all".

Same setting in `robotiq_controllers.sim.humble.yaml`. Since velocity is in `state_interfaces` here, consider leaving `allow_stalling: false` for the sim configs, or documenting that the simulator must publish real velocities for stall detection to mean anything.

README.md
```diff
@@ -256,10 +256,25 @@ ros2 launch robotiq_description robotiq_control.launch.py                    # r
...
 
+`sim_gazebo` and `sim_isaac` swap the hardware plugin for `gz_ros2_control/GazeboSimSystem` or
+`topic_based_ros2_control/TopicBasedSystem` (any simulator that exchanges `sensor_msgs/JointState`
+on two topics — Isaac Sim is the usual one, hence the argument names). Both plugins export only the
+standard joint interfaces, so the launch loads `config/robotiq_controllers.sim.yaml` for them
+(per-goal `effort`/`velocity` are accepted and ignored; stall detection is tuned for a loop that
+runs over a topic) and skips `robotiq_activation_controller`, whose `reactivate_gripper` GPIO only
+the driver and the mock declare. `TopicBasedSystem` matches joints to the simulator's `JointState`
+**by name**, so the simulator must publish this description's joint names (`robotiq_85_left_knuckle_joint`
+and the five mimic joints, with your `prefix`), or nothing moves. Neither plugin is a dependency of
```
mbegin-robotiq: This note lists only the 2F-85 joint names, but the launch supports 2F-140 as well. A 2F-140 Isaac user who follows it verbatim publishes `robotiq_85_*` names, nothing matches, and nothing moves — with no error to explain why. Either give both name sets or phrase it in terms of the model's prefix/joint pattern.

mbegin-robotiq (approved):
Approving. The Isaac/fake paths and the config-selection plumbing look right, and the tests cover them. The four inline findings above — the `sim_gazebo` path in particular — are worth addressing in a follow-up rather than blocking this.

grippers/robotiq_description/launch/robotiq_control.launch.py
jlongvalRobotiq: Dropped in 06cd264: the launch argument, its forwarding to xacro, the top-level xacro arg and the README line are gone; only `sim_isaac` remains, which is the path verified live. The macro-level `sim_gazebo` parameter predates this PR and stays for cells that embed the gripper macro in their own Gazebo description; the README now says so in one sentence. With one flag left, the two `PythonExpression` evals went too (config choice is a small `Substitution` on launch's own condition parser).

grippers/robotiq_description/launch/robotiq_control.launch.py
jlongvalRobotiq: 610ec61 adds a check that runs before any node starts and fails the launch with the offending names, e.g. `use_fake_hardware and sim_isaac both select a hardware plugin; set at most one` (verified live). Tests cover the three accepted combinations and the rejected one.

grippers/robotiq_description/config/robotiq_controllers.sim.yaml
jlongvalRobotiq: Set `allow_stalling: false` in both sim configs in 4a90921, so a stall aborts instead of succeeding; the client decides what an aborted, stalled goal means. Kept `stall_timeout` / `stall_velocity_threshold`: the detector runs either way and those decide when the abort fires, and the upstream default threshold (0.001 rad/s) sits under simulated velocity noise.

README.md
jlongvalRobotiq: Replaced with a table in 38bf816: driven joint and the five mimic joints for each `gripper_model`.

grippers/robotiq_description/config/robotiq_controllers.sim.humble.yaml
```diff
@@ -0,0 +1,26 @@
...
+    robotiq_gripper_controller:
+      type: position_controllers/GripperActionController
+
+robotiq_gripper_controller:
+  ros__parameters:
+    joint: $(var gripper_joint)
+
+    max_effort: 50.0
+
+    allow_stalling: false
```
ebarnett3, line 23: (relates to robotiq/ros#29): the stall
detector only looks at the velocity state. Per the Jazzy controller source,
each cycle: inside `goal_tolerance` succeeds; otherwise |velocity| above
`stall_velocity_threshold` resets the timer, and once `stall_timeout`
elapses with no movement the goal is marked stalled, then SUCCEEDED if
`allow_stalling` else ABORTED.

So on a simulator that omits the `JointState` velocity array, every goal
longer than 0.5 s fails, including goals that would have arrived: #29's
false stall with a harsher outcome. Isaac Sim's joint-state publisher
normally includes velocities, which is why the author's live test passed;
whether the twin in use does was not confirmed before merge. 

This is a regression that will be fixed with the changes for https://github.com/robotiq/ros/issues/63

grippers/robotiq_description/config/robotiq_controllers.sim.yaml
```diff
@@ -0,0 +1,40 @@
...
+#   standard position/velocity/effort interfaces only, never the driver's
+#   `set_gripper_max_*`, and a claimed interface that does not exist stops the
+#   controller from activating. Goal effort/velocity are accepted and ignored.
+# - A stall aborts the goal instead of succeeding. A simulator that publishes no
+#   joint velocities leaves the velocity state at 0, which reads as "stalled"
+#   from the first cycle, so allow_stalling would report every out-of-tolerance
+#   goal as succeeded. The client decides whether an aborted, stalled goal means
+#   an object was gripped. The stall window itself is wider than the driver's:
+#   a goal needs longer than 0.05 s to show up as motion over a topic, and
+#   simulated joint velocities carry more noise than the driver's.
```
ebarnett3, line 23: The driver publishes no joint velocity at all; its state is
written as 0 every cycle (#29). There is no driver velocity to be quieter than,
and `robotiq_controllers.yaml:30`'s `stall_timeout: 0.05` is not a tuned
reference, it is the value under which #29's false stall fires 50 ms after
every goal. The 0.5 s / 0.05 rad/s in the sim config should be justified
against the topic-fed (or relay-estimated) velocity on its own terms, and the
hardware values re-chosen when #29 lands.

grippers/robotiq_description/config/robotiq_controllers.sim.humble.yaml
```diff
@@ -0,0 +1,26 @@
+# Humble controller configuration for the simulated hardware plugin
```
ebarnett3, line 1: I find it confusing to refer to a "simulated hardware plugin" if it always refers specifically to isaac_sim. Same thing with the file name robotiq_controllers.sim.humble.yaml

With this terminology, a reader could easily think we are referring to the mock gripper in ros.

grippers/robotiq_description/config/robotiq_controllers.sim.humble.yaml
ebarnett3: Clarification: under
`use_fake_hardware:=true` the mock teleports (GenericSystem copies the position
command into the state each cycle, and the controller writes the goal straight
to the command), so a goal succeeds on the next update, velocity is
meaningless and stall detection never runs. That path keeps the hardware
config and is unaffected by #29 or #63. Under `sim_isaac:=true` the simulator
integrates the joint with its own drives, the finger moves over time, and the
velocity the sim config's stall parameters read is a real quantity. The two
should not be lumped together as "simulated".

README.md
```diff
@@ -257,10 +257,33 @@ ros2 launch robotiq_description robotiq_control.launch.py                    # r
...
+joints to the simulator's `JointState` **by name**, so the simulator must publish this description's
+six joint names for the model you launch, each with your `prefix`, or nothing moves:
+
+| `gripper_model` | driven joint | mimic joints |
+|---|---|---|
+| `2f_85` | `robotiq_85_left_knuckle_joint` | `robotiq_85_right_knuckle_joint`, `robotiq_85_left_inner_knuckle_joint`, `robotiq_85_right_inner_knuckle_joint`, `robotiq_85_left_finger_tip_joint`, `robotiq_85_right_finger_tip_joint` |
+| `2f_140` | `finger_joint` | `right_outer_knuckle_joint`, `left_inner_knuckle_joint`, `right_inner_knuckle_joint`, `left_inner_finger_joint`, `right_inner_finger_joint` |
+
+The plugin is not a dependency of this package — build
+[topic_based_ros2_control](https://github.com/PickNikRobotics/topic_based_ros2_control) yourself.
```
ebarnett3, line 282: this does not say the build is only
needed on Jazzy and Lyrical. rosdistro releases the plugin as a binary for
Humble only (`ros-humble-topic-based-ros2-control`); Jazzy, Lyrical and
Rolling have no release. Suggested replacement:

    The plugin is not a dependency of this package. On Humble install
    `ros-humble-topic-based-ros2-control`; on Jazzy and Lyrical it has no
    binary release, so build it in an overlay once:
      git clone https://github.com/PickNikRobotics/topic_based_ros2_control <ws>/src/
      colcon build --cmake-args -DBUILD_TESTING=OFF
    (`-DBUILD_TESTING=OFF` skips its `ros_testing` test dependency, which a
    runtime install does not have.)

PR #55, opened 2026-09-03
build(mcp): add the gripper MCP server skeleton

jlongvalRobotiq:
## Problem

The MCP server that exposes the 2F grippers to LLM agents lives in its own repo with every change pushed straight to main, so nothing there is reviewable, and it duplicates knowledge (joint names, action surface, stall semantics) that belongs next to the driver it drives.

## Change

This is the first of a series of small PRs that rebuild it inside this repo as `mcp/`. It adds the skeleton only:

1. **`mcp/`** — a plain uv Python project (`robotiq-gripper-mcp`, import package `gripper_mcp`), Python 3.12, `COLCON_IGNORE` so colcon never sees it. It cannot be an ament package: fastmcp and pydantic v2 have no rosdep keys, and fastmcp needs 3.12, which Humble does not ship. The README says so. No runtime dependencies yet; they arrive with the PRs that use them.
2. **CI** — `ci-mcp.yml` runs `uv sync --locked` + pytest on 3.12, path-filtered on `mcp/**`. The repo-wide pre-commit config and `ci-format` now cover `mcp/` too.
3. **Versioning** — `mcp/pyproject.toml` is one more literal copy of the repo version, so `dev/version.py` now propagates and checks it alongside the `package.xml` files, `VERSION` lists it, and the pre-commit hook fires on `pyproject.toml` edits.

Follow-ups, in order: units and response models; backend protocol + mock; gripper datasheets + config; service layer; MCP tool surface; ROS 2 backend on rclpy (`gripper_cmd` action); tactile protocol + mock; tactile ROS backend; Docker image.

## Verification

- `uv run pytest`: 1 passed (import smoke test, so the job is green and non-empty).
- `dev/version.py --check`: 6 copies agree; setting the pyproject to 9.9.9 fails the check with the file named.


mbegin-robotiq:
@jlongvalRobotiq ready for review ?

mcp/README.md
```diff
@@ -0,0 +1,36 @@
...
+requests.
+
+## Not a ROS package
+
+`mcp/` is a plain Python project managed with [uv](https://docs.astral.sh/uv/),
+not an ament package, and `COLCON_IGNORE` keeps it out of every colcon build:
+
+- its dependencies (fastmcp, pydantic v2) have no rosdep keys, so a `package.xml`
+  could not declare them;
+- fastmcp needs Python 3.12, which Humble (Python 3.10) does not ship.
```
ebarnett3: ###  "fastmcp needs Python 3.12" is not what fastmcp declares — Low (doc)

`mcp/README.md:20`, `.github/workflows/ci-mcp.yml:3` and the PR body all justify
the non-ament choice with "fastmcp needs 3.12, which Humble does not ship".
fastmcp's metadata on PyPI (4.0.3 today) says `requires_python >= 3.10`, and the
mcp SDK is `>= 3.10` too. The rosdep argument on `mcp/README.md:18-19` is
sufficient and true on its own; the Humble line is false as written.

Fix: either drop the second bullet, or say what is actually true:

```markdown
- the project targets Python 3.12 (Jazzy's interpreter) and is only tested there;
  a Humble host would have to run it from a separate 3.12 install.
```

and mirror the change in the `ci-mcp.yml` header comment.

mcp/uv.lock
```diff
@@ -0,0 +1,79 @@
...
+]
+sdist = { url = "https://files.pythonhosted.org/packages/e4/47/b9efed96c114afcfa3c9d3fe98a76a1d14c74a9e266d397cf6eb64be5e01/pytest-9.1.1.tar.gz", hash = "sha256:1088fbde8f2b49d95a549a195707afa7a76a3ce9bcadc26b6d71f0ffda5fe313", size = 1636369, upload-time = "2026-06-19T10:58:32.857Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/24/25/1de2678b631f5a49215c6c96fff41ba892b0a34df68d6d80292b1b48aa7f/pytest-9.1.1-py3-none-any.whl", hash = "sha256:37a86b45efb9a47a61a36449063e8e18d0cab3161329fc099eb21783169c4f0c", size = 386536, upload-time = "2026-06-19T10:58:31.347Z" },
+]
+
+[[package]]
+name = "robotiq-gripper-mcp"
+version = "1.1.0"
+source = { editable = "." }
```
ebarnett3, line 83: ###  `dev/version.py --set` leaves `mcp/uv.lock` stale, so the next release bump breaks the MCP CI job — Medium

`mcp/uv.lock:66-69` records the root project's own version:

```
[[package]]
name = "robotiq-gripper-mcp"
version = "1.1.0"
source = { editable = "." }
```

`dev/version.py:30` only knows `package.xml` and `pyproject.toml`, and
`.github/workflows/ci-mcp.yml:32` runs `uv sync --locked`. Reproduced on the
tip of the train: bump the pyproject to 1.2.0 and

```
$ uv lock --locked
The lockfile at `uv.lock` needs to be updated, but `--locked` was provided.
```

So the first `dev/version.py --set X.Y.Z` after this merges passes the
pre-commit hook and fails the `MCP` workflow on the release PR.

Fix: treat the lock as one more copy. In `dev/version.py:29-30`:

```python
PYPROJECT_VERSION = re.compile(r'^(version = ")([^"]*)(")', re.MULTILINE)
UV_LOCK_VERSION = re.compile(r'(name = "robotiq-gripper-mcp"\nversion = ")([^"]*)(")')
COPIES = {
    "package.xml": VERSION_TAG,
    "pyproject.toml": PYPROJECT_VERSION,
    "uv.lock": UV_LOCK_VERSION,
}
```

Everything else already generalises (`ls-files "*uv.lock"`, `endswith(tuple(COPIES))`,
`count=1`). Then regenerate the list with `dev/version.py --set 1.1.0`, add
`mcp/uv.lock` to the reminder block in `VERSION`, and widen the hook trigger in
`.pre-commit-config.yaml:81` to `'(^VERSION$|package\.xml$|pyproject\.toml$|uv\.lock$)'`.
Alternative if you prefer uv to own its file: have `--set` run
`uv lock --project mcp` after rewriting the pyproject, but that puts a uv
dependency in a repo-wide dev script, so the regex is the smaller change.

mcp/gripper_mcp/__init__.py
```diff
@@ -0,0 +1 @@
+"""MCP server for Robotiq 2F adaptive grippers, on the ROS 2 driver in this repo."""
```
ebarnett3, line 1: Does it support Hand-E ?

mcp/uv.lock
```diff
@@ -0,0 +1,79 @@
...
+
+[package.dev-dependencies]
+dev = [
+    { name = "pytest" },
+]
+
+[package.metadata]
+
+[package.metadata.requires-dev]
+dev = [{ name = "pytest", specifier = ">=8.0" }]
```
ebarnett3: ### README does not say how `uv.lock` is maintained — Suggestion

`mcp/README.md:27-32` shows `uv sync` and `uv run pytest` but never mentions the
lock, so the rules (commit it, how deps get bumped, why CI uses `--locked`)
are implicit. Proposed addition under "Development":

```markdown
`uv.lock` is generated from `pyproject.toml` and committed: it pins every
dependency to an exact file and hash, and CI and the Docker image install
from it with `uv sync --locked`, which fails if it is stale.

- `uv add <pkg>` / `uv remove <pkg>` edit `pyproject.toml` and the lock
  together; plain `uv sync` also refreshes the lock after a hand edit.
- `uv lock --upgrade` (or `--upgrade-package <pkg>`) moves pinned versions
  within the ranges `pyproject.toml` declares. Nothing else does.
- The lock also records this project's own version, so a release bump
  regenerates it (see `dev/version.py`).
```

We could also make it more obvious that the file is generated by creating a new file `.gitattributes` a the repo root with the content
```
mcp/uv.lock linguist-generated=true
```

ebarnett3 (approved):

mcp/gripper_mcp/__init__.py
jlongvalRobotiq: Not today. The server drives the ROS driver in `grippers/`, and `robotiq_description` only ships URDFs and controller configs for the 2F-85 and 2F-140. On the MCP side a model is one `robot_specifications/<model>.yaml`: command joint name, its open/closed range from the URDF, and the stroke in mm. Hand-E would be that one file the day the description package carries it.

mcp/README.md
jlongvalRobotiq: You were right, it was a choice, not a constraint. 60aba1d declares `>=3.10` and CI now runs pytest on 3.10 and 3.12; README rewritten accordingly. The only Humble blocker left is a top-level import in the ROS backend, handled in #61.

mcp/uv.lock
jlongvalRobotiq: Done in c310857, your regex approach: uv.lock is a COPIES entry, the hook trigger is widened, VERSION list regenerated. Verified `--set 1.2.0` then `uv lock --locked` passes.

mcp/uv.lock
jlongvalRobotiq: Added in c310857, your paragraph plus the .gitattributes line.

PR #56, opened 2026-09-03
feat(mcp): opening/joint-angle conversions and response models

jlongvalRobotiq:
## Change

Second PR of the `mcp/` series (stacked on #55). Two pure modules, no I/O, no ROS:

1. **`units.py`** — the conversion between the opening an agent asks for (mm, 0 = closed) and the position `robotiq_gripper_controller` wants for the command joint (rad, 0 = open, inverted). Linear, with clamping at both ends of the stroke. The real linkage is a four-bar, so mid-stroke openings are approximate; the module docstring says so. The `GripperGeometry` dataclass holds the four numbers per model; the per-model values arrive with the datasheets PR.
2. **`models.py`** — the pydantic response models the tools will return: `GripperInfo`, `GripperState`, `GripperMotionResult`, `GripperHealth`. Every model names its units and carries the `backend` (`ros` or `mock`) that produced it. `Outcome` and `Backend` are closed literals.

Not carried over from gripper-mcp, deliberately: the SDK register domain (rPR/gPO) and percent-to-byte helpers, and the SDK-only state fields (motor current, gOBJ, gSTA, gFLT). This server drives the ROS driver, which exposes none of those. Tactile models come with the tactile tools.

`pydantic>=2` is the first runtime dependency.

## Verification

- `uv run pytest`: 14 passed. Units: both stroke ends, mid-stroke fraction, clamping in both directions, round-trip over five openings. Models: unknown backend and unknown outcome are rejected, a result survives a JSON round-trip.

## Stack
Stacked on #55: this PR shows only its own 2 commits.


mcp/pyproject.toml
```diff
@@ -5,7 +5,9 @@ description = "MCP server exposing Robotiq 2F grippers to LLM agents through the
 readme = "README.md"
 license = "BSD-3-Clause"
 requires-python = ">=3.12"
-dependencies = []
+dependencies = [
+    "pydantic>=2.0",
```
jlongvalRobotiq: should we fix a value here ? >= 2.0 seems like a good way to introduce bugs ? Same idea for all other dependancies in the project.

ebarnett3: I agree that a fixed version is better for stability.

mcp/gripper_mcp/units.py
```diff
@@ -0,0 +1,50 @@
...
+    fraction = opening_mm_to_fraction(opening_mm, geometry)
+    span = geometry.knuckle_rad_closed - geometry.knuckle_rad_open
+    return geometry.knuckle_rad_closed - fraction * span
+
+
+def knuckle_rad_to_opening_mm(knuckle_rad: float, geometry: GripperGeometry) -> float:
+    span = geometry.knuckle_rad_closed - geometry.knuckle_rad_open
+    fraction = (geometry.knuckle_rad_closed - knuckle_rad) / span
+    opening_span = geometry.max_opening_mm - geometry.min_opening_mm
+    return geometry.min_opening_mm + clamp(fraction, 0.0, 1.0) * opening_span
```
ebarnett3, line 61: ### The mm conversion is a second conversion layer, in a second repo, on top of the driver's — Design

`mcp/gripper_mcp/units.py` maps mm to the command joint's radians; the driver
in this repo already maps those radians to register counts
(`grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp:58-78`),
with a comment that the joint unit is "whatever the URDF uses". Two linear maps
in two languages, each with its own idea of "closed" (#58 finding 1 shows them
already disagreeing). All conversions should live in one place, and the only
place the MCP can reach over ROS is the joint unit itself.

Recommended, tracked in robotiq/grippers#16: the general conversion,
counts to opening in mm, belongs in the C++ SDK. It depends only on the
gripper: stroke per model and the usable register band. Every consumer wants
it and none of it is ROS or MCP specific. The MCP would access the C++ SDK
conversions through ROS.

Until then, for this PR: keep `units.py` but say in the module docstring that
it is a stopgap pending robotiq/grippers#16, and take its constants from
`/robot_description`, so the interim
code is small and holds no second copy of the calibration.

mcp/gripper_mcp/models.py
```diff
@@ -0,0 +1,66 @@
...
+    "grasped",
+    "stalled_unexpectedly",
+    "incomplete",
+    "not_supported",
+    "refused",
+]
+
+
+class GripperInfo(BaseModel):
+    name: str
```
ebarnett3: `mcp/gripper_mcp/models.py:25` calls the gripper `name`; `GripperState`,
`GripperMotionResult` and `GripperHealth` (lines 32, 47, 60) call the same thing
`robot_name`, and `grippers.yaml.example:2` in #58 tells the user that `name` is
"what agents pass as robot_name". An agent listing grippers and then addressing
one has to know that `info.name` becomes `robot_name=`. Pick one before any tool
ships it. Since the tool parameter is `robot_name`:

```python
class GripperInfo(BaseModel):
    robot_name: str
    model: str
    backend: Backend
    description: str
```

mcp/gripper_mcp/units.py
```diff
@@ -0,0 +1,50 @@
...
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class GripperGeometry:
+    max_opening_mm: float
+    min_opening_mm: float
+    knuckle_rad_open: float
+    knuckle_rad_closed: float
```
ebarnett3, line 29: ###  `GripperGeometry` accepts a zero span and fails later with `ZeroDivisionError` — Medium

`mcp/gripper_mcp/units.py:20-24` is a plain frozen dataclass; `units.py:36` and
`:49` divide by `max_opening_mm - min_opening_mm`, and `:42`/`:47` by
`knuckle_rad_closed - knuckle_rad_open`. #58 loads this dataclass straight from
YAML through pydantic, which validates field types but not the invariant, so a
datasheet typo (`max_opening_mm: 0.0`) passes config loading and blows up on the
first `read_state`. Check it where the numbers are born:

```python
@dataclass(frozen=True)
class GripperGeometry:
    max_opening_mm: float
    min_opening_mm: float
    knuckle_rad_open: float
    knuckle_rad_closed: float

    def __post_init__(self) -> None:
        if self.max_opening_mm <= self.min_opening_mm:
            raise ValueError("max_opening_mm must exceed min_opening_mm")
        if self.knuckle_rad_closed == self.knuckle_rad_open:
            raise ValueError("knuckle_rad_closed must differ from knuckle_rad_open")
```

pydantic runs `__post_init__` when validating a stdlib dataclass, so #58's
"missing field is rejected" test gets a sibling for free.

mcp/gripper_mcp/units.py
```diff
@@ -0,0 +1,50 @@
...
+
+- **opening_mm**: what agents and tools speak. 0.0 = closed, the model's max
+  opening (85.0 on a 2F-85) = fully open.
+- **knuckle_rad**: what `robotiq_gripper_controller` speaks, the position of the
+  gripper's command joint. 0.0 = open, the joint's upper limit (0.8 on a 2F-85)
+  = closed. INVERTED relative to opening_mm.
+
+The mapping between them is linear. The real linkage is a four-bar, so the
+opening this yields is an approximation, good to a couple of millimetres at
+mid-stroke and exact at both ends.
```
ebarnett3: ###  "good to a couple of millimetres at mid-stroke" is unsourced, and "exact at both ends" is false with #58's numbers — Low (doc)

The closed end is off by 0.75 mm on the 85 because #58 uses the URDF limit
(0.8) where the driver reports `gripper_closed_position` (0.7929).

`mcp/gripper_mcp/units.py:11-13` states a numeric accuracy for the linear map.
Nothing in the PR measures it, and the linkage is not modelled anywhere. A crude
single-pivot estimate for the 2F-85 (knuckle length chosen so 0.8 rad closes
85 mm) puts the mid-stroke error at 3-4 mm, but that is a guess too. Either
measure it against the driver's reported position on a real 2F-85 and cite the
number, or drop the figure:

```
The mapping between them is linear. The real linkage is a four-bar, so
mid-stroke openings are approximate; both ends are exact.
```

mcp/gripper_mcp/models.py
```diff
@@ -0,0 +1,66 @@
...
+    opening_mm: float = Field(description="0.0 = closed, the model's max = fully open")
+    opening_fraction: float = Field(description="0.0 = closed, 1.0 = fully open")
+    knuckle_rad: float = Field(
+        description="Command joint position, controller convention: 0.0 = open"
+    )
+    force_n: float | None = Field(
+        default=None, description="Measured grip force in newtons; null when unmeasured"
+    )
+    backend: Backend
+    measured_at: str = Field(description="ISO-8601 UTC")
```
ebarnett3: ### `measured_at: str` should be a `datetime` — Low

`mcp/gripper_mcp/models.py:42`. pydantic v2 serialises an aware `datetime` as
ISO-8601 with offset and validates it on the way back in, which is exactly what
the description promises; a `str` promises nothing.

```python
from datetime import datetime
...
    measured_at: datetime = Field(description="UTC, timezone-aware")
```

Producers then use `datetime.now(timezone.utc)`.

mcp/gripper_mcp/units.py
```diff
@@ -0,0 +1,50 @@
+"""Conversions between the agent-facing opening (mm) and the controller's joint angle.
+
+Two domains are in play and they must never be confused:
+
+- **opening_mm**: what agents and tools speak. 0.0 = closed, the model's max
+  opening (85.0 on a 2F-85) = fully open.
+- **knuckle_rad**: what `robotiq_gripper_controller` speaks, the position of the
+  gripper's command joint. 0.0 = open, the joint's upper limit (0.8 on a 2F-85)
+  = closed. INVERTED relative to opening_mm.
```
ebarnett3, line 9: ### `GripperState` exposes both domains the module docstring says must never be confused — Question

`units.py:3-9` insists the agent speaks mm and never the controller's inverted
radians, yet `models.py:35` returns `knuckle_rad` to the agent alongside
`opening_mm` and `opening_fraction`. If it is there for debugging, say so in
the field description ("controller-side value, for diagnostics; use
`opening_mm`"), otherwise drop it. As written an LLM will happily feed it back
into `gripper_move(opening_mm=...)`.

ebarnett3 (approved):

mcp/pyproject.toml
jlongvalRobotiq: Pinned: pydantic==2.13.5 here (folded into the commit that adds it), pytest in #55, pyyaml and fastmcp in the PRs that add them.

mcp/gripper_mcp/units.py
jlongvalRobotiq: Agreed on the destination; the module docstring now says units.py is a stopgap pending robotiq/grippers#16 (3fa4a4c). I kept the datasheet as the source of the constants rather than `/robot_description`: the mock backend has no graph to read from, and the datasheet values are the URDF's already, so it would add a ROS dependency to the conversion without removing a copy. The 0.7929 vs 0.8 closed value gets fixed in #58, where the datasheet lives.

mcp/gripper_mcp/models.py
jlongvalRobotiq: Renamed to `robot_name` in 223b45d.

mcp/gripper_mcp/units.py
jlongvalRobotiq: Added in 3fa4a4c, your `__post_init__`, with two tests.

mcp/gripper_mcp/units.py
jlongvalRobotiq: Dropped in 3fa4a4c, your wording.

mcp/gripper_mcp/models.py
jlongvalRobotiq: `datetime` in 223b45d, with a wire round-trip test that checks the offset survives.

mcp/gripper_mcp/units.py
jlongvalRobotiq: Kept for diagnostics; the description now says so and points at `opening_mm` (223b45d).

PR #57, opened 2026-09-03
feat(mcp): backend protocol and mock backend

jlongvalRobotiq:
## Change

Third PR of the `mcp/` series (stacked on #56). The seam between the tool surface and whatever moves the fingers:

1. **`backend.py`** — the `GripperBackend` protocol and its three result dataclasses. Backends speak the command joint's position in radians and nothing else; the service layer (next PR but one) owns the millimetre conversion and the pydantic models. The vocabulary is the driver's `gripper_cmd` action: position + max effort in, `reached_goal` + `stalled` out. A stall is passed through untouched; deciding whether it was a wanted grasp is the caller's job.
2. **`mock_backend.py`** — the CI default. A virtual object of a given width sits between the fingers: closing past it stalls there and the commanded effort becomes the reported grip force. Travel time comes from an injected sleep so tests run instantly, and a goal slower than its timeout comes back `timed_out` with the fingers where they were.

Not carried over from gripper-mcp, deliberately: the speed byte, activation, fault codes and register domain (the `gripper_cmd` action exposes none of them), and `close()` on the protocol (tearing down an rclpy node belongs to whoever builds the ROS backend, not to the service-facing interface).

## Verification

- `uv run pytest`: 25 passed (11 new). Mock: satisfies the protocol, starts open, reaches the goal on empty space, stalls at the object width with the effort as grip force, a narrower object does not stall, opening after a grasp releases, targets past the stroke clamp, a slow move times out without moving, travel time grows with distance.

## Stack
Stacked on #56: this PR shows only its own 2 commits.

### DIFF
--- a/mcp/gripper_mcp/backend.py
+++ b/mcp/gripper_mcp/backend.py
@@ -1,0 +1,52 @@
+"""What a gripper backend has to provide.
+
+Backends speak the controller's domain, the command joint's position in radians
+(0 = open), and know nothing about millimetres, pydantic or MCP; the service
+layer owns that translation. This is what lets the same tool surface sit on the
+ROS driver, real or simulated, or on a mock without any of them leaking into the
+tool signatures.
+
+The vocabulary is the `gripper_cmd` action's: a position and a max effort in,
+`reached_goal` and `stalled` out. A stall on a close is how the driver reports
+an object between the fingers, so backends pass it through untouched and never
+decide for the caller whether it was wanted.
+"""
+
+from dataclasses import dataclass
+from typing import Protocol, runtime_checkable
+
+
+@dataclass(frozen=True)
+class BackendState:
+    position_rad: float
+    force_n: float | None = None
+
+
+@dataclass(frozen=True)
+class BackendMotion:
+    final_position_rad: float
+    reached_goal: bool
+    stalled: bool
+    refused: bool = False
+    timed_out: bool = False
+    detail: str = ""
+
+
+@dataclass(frozen=True)
+class BackendHealth:
+    reachable: bool
+    controller_active: bool | None = None
+    detail: str = ""
+
+
+@runtime_checkable
+class GripperBackend(Protocol):
+    name: str
+
+    def read_state(self) -> BackendState: ...
+
+    def move_to(
+        self, position_rad: float, max_effort_n: float, timeout_s: float
+    ) -> BackendMotion: ...
+
+    def health(self) -> BackendHealth: ...
--- a/mcp/gripper_mcp/backend.py
+++ b/mcp/gripper_mcp/backend.py
@@ -3,13 +3,20 @@
 Backends speak the controller's domain, the command joint's position in radians
 (0 = open), and know nothing about millimetres, pydantic or MCP; the service
 layer owns that translation. This is what lets the same tool surface sit on the
-ROS driver, real or simulated, or on a mock without any of them leaking into the
-tool signatures.
+ROS driver, real or simulated, or on a test double without any of them leaking
+into the tool signatures.
 
 The vocabulary is the `gripper_cmd` action's: a position and a max effort in,
 `reached_goal` and `stalled` out. A stall on a close is how the driver reports
 an object between the fingers, so backends pass it through untouched and never
-decide for the caller whether it was wanted.
+decide for the caller whether it was wanted. `refused` is set when the backend
+rejected the goal before any motion, typically the action server turning it
+down; `timed_out` when the motion outlived its deadline.
+
+This contract is not MCP-specific: every Python client of the ROS driver wants
+"move to a position with a max effort, report reached or stalled". Its home is
+a ros client package alongside the ROS backend once that backend lands; it sits
+here until then, with the MCP as its first consumer.
 """
 
 from dataclasses import dataclass
--- a/mcp/gripper_mcp/mock_backend.py
+++ b/mcp/gripper_mcp/mock_backend.py
@@ -1,0 +1,120 @@
+"""Pure-Python mock gripper, the CI default backend.
+
+It exists to exercise the behaviours the service layer has to get right and a
+unit test cannot reach through the real driver: a close that stalls on an
+object, travel time, and a goal that outlives its timeout. An optional virtual
+object of a given width sits between the fingers; closing past it stalls there.
+
+Travel is simulated through an injected `sleep_fn` so tests run instantly while
+a container still moves in something like real time.
+"""
+
+import time
+from typing import Callable
+
+from gripper_mcp.backend import BackendHealth, BackendMotion, BackendState
+from gripper_mcp.units import (
+    GripperGeometry,
+    clamp,
+    knuckle_rad_to_opening_mm,
+    opening_mm_to_knuckle_rad,
+)
+
+MOCK_GEOMETRY = GripperGeometry(
+    max_opening_mm=85.0,
+    min_opening_mm=0.0,
+    knuckle_rad_open=0.0,
+    knuckle_rad_closed=0.8,
+)
+
+NOMINAL_TRAVEL_SPEED_MM_S = 150.0
+
+
+class MockGripperBackend:
+    name = "mock"
+
+    def __init__(
+        self,
+        object_width_mm: float | None = None,
+        travel_speed_mm_s: float = NOMINAL_TRAVEL_SPEED_MM_S,
+        sleep_fn: Callable[[float], None] = time.sleep,
+        geometry: GripperGeometry = MOCK_GEOMETRY,
+    ) -> None:
+        self._object_width_mm = object_width_mm
+        self._travel_speed_mm_s = travel_speed_mm_s
+        self._sleep = sleep_fn
+        self._geometry = geometry
+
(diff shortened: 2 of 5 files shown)


mcp/gripper_mcp/backend.py
```diff
@@ -0,0 +1,52 @@
+"""What a gripper backend has to provide.
+
+Backends speak the controller's domain, the command joint's position in radians
+(0 = open), and know nothing about millimetres, pydantic or MCP; the service
+layer owns that translation. This is what lets the same tool surface sit on the
+ROS driver, real or simulated, or on a mock without any of them leaking into the
+tool signatures.
```
mbegin-robotiq: if we are ROS-agnostic and MCP agnostic (which is great) then the backend probably belongs in the SDK repo ?

jlongvalRobotiq: I'd keep it here: this file is what the MCP service expects from a backend (position and effort in, `reached_goal`/`stalled` out), so the MCP owns the contract. Backends are implementations of it and can live anywhere. Since it's a `typing.Protocol`, a future Modbus backend on the C++ SDK satisfies it just by having the same methods, without importing this file.

mcp/tests/fakes/gripper.py
```diff
@@ -0,0 +1,120 @@
...
+            detail=f"Mock gripper at {opening_mm:.1f} mm.",
+        )
+
+    def _resolve_stop(self, target: float) -> tuple[float, bool]:
+        closing = target > self._position_rad
+        if not closing or self._object_width_mm is None:
+            return target, False
+
+        obstruction = self.position_rad_for(self._object_width_mm)
+        if target <= obstruction:
```
ebarnett3, line 132: ### 2. A virtual object wider than the stroke "stalls" a close that never moved — Low (only if the shipped mock survives finding 1)

`mock_backend.py:117` converts `object_width_mm` through the clamping
`opening_mm_to_knuckle_rad`, so `object_width_mm=120` on a 2F-85 becomes an
obstruction at 0.0 rad. Any close then returns `stalled=True`, `force_n =
max_effort_n`, with the fingers fully open. The real gripper cannot stall
before it starts moving. Refuse the configuration instead:

```python
        if object_width_mm is not None and not (
            geometry.min_opening_mm < object_width_mm < geometry.max_opening_mm
        ):
            raise ValueError(
                f"object_width_mm={object_width_mm} is outside the stroke "
                f"[{geometry.min_opening_mm}, {geometry.max_opening_mm}] mm"
            )
```

mcp/gripper_mcp/mock_backend.py
```diff
@@ -0,0 +1,120 @@
...
+unit test cannot reach through the real driver: a close that stalls on an
+object, travel time, and a goal that outlives its timeout. An optional virtual
+object of a given width sits between the fingers; closing past it stalls there.
+
+Travel is simulated through an injected `sleep_fn` so tests run instantly while
+a container still moves in something like real time.
+"""
+
+import time
+from typing import Callable
```
ebarnett3: ### 3. `from typing import Callable` — Nit (only if the shipped mock survives finding 1)

`mock_backend.py:13`. Deprecated alias since 3.9; the project is 3.12-only.

```python
from collections.abc import Callable
```

mcp/gripper_mcp/backend.py
```diff
@@ -0,0 +1,52 @@
...
+    position_rad: float
+    force_n: float | None = None
+
+
+@dataclass(frozen=True)
+class BackendMotion:
+    final_position_rad: float
+    reached_goal: bool
+    stalled: bool
+    refused: bool = False
```
ebarnett3, line 37: ### 4. `BackendMotion.refused` has no producer in this PR — Question

`backend.py:30`. Neither the mock nor the protocol docstring says when a backend
sets it (goal rejected by the action server, presumably). One sentence in the
dataclass would save the #59 reader a trip to the ROS backend:

```python
    refused: bool = False  # the backend rejected the goal before moving (e.g. action server refused it)
```

mcp/gripper_mcp/backend.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+@runtime_checkable
+class GripperBackend(Protocol):
+    name: str
+
+    def read_state(self) -> BackendState: ...
+
+    def move_to(
+        self, position_rad: float, max_effort_n: float, timeout_s: float
```
ebarnett3, line 56: ### 1. Only MCP-specific code belongs in `mcp/`; the backend contract, its ROS implementation and the mock are not — Design (decide before merging)

Sorting this PR and its follow-up #61 by who wants the code:

- `backend.py`: the contract "move to a position with a max effort, report
  reached or stalled; read the joint; health" is what every Python user of the
  ROS driver wants, and #61's `RosGripperBackend` (action client on
  `gripper_cmd`, `joint_states` subscriber, health probe) is the implementation
  every one of them will otherwise rewrite. The ros repo has no Python client
  today. Both belong there, as an ament_python package (say
  `grippers/robotiq_gripper_client`), built and tested in the ROS CI matrix,
  with the MCP as its first consumer. Redefining `move_to` in each consumer is
  the debt this repo is trying to retire.
- `mock_backend.py`: a fake gripper is already provided twice below the MCP,
  and the SDK's is the one to use. `Robotiq::makeFakeGripper()` is an
  in-process fake device behind the real `Gripper`, and `robotiq_driver`
  already wraps it: the `use_dummy` hardware parameter makes the real plugin
  drive the fake (`hardware_interface.cpp:149`). Pointing the MCP's ros backend
  at a controller manager running that way exercises the driver, the
  controller and the action, which a Python mock never will.
  (`use_fake_hardware:=true` is the other path, ros2_control's generic mock; it
  bypasses the driver, so it is the weaker choice here.) A Python re-model of
  a gripper with a virtual object is a third fake, owned by the one consumer
  that cannot share it. Two follow-ups, both outside this PR:
  robotiq/ros: expose `use_dummy` from `2f_common.ros2_control.xacro:25` (hard
  coded `false` today) as a launch argument next to `use_fake_hardware`.
  robotiq/grippers: the fake models no travel and no object detection
  (`src/fake/register_model.hpp:11-15`), by design. The behaviour this mock
  reaches for (constant-speed travel, stall on an object, status sequencing)
  is what a `makeSimulatedGripper()` in the SDK should provide, once, for the
  driver, the MCP, and the examples (issue to be filed on
  robotiq/grippers).
- What is MCP-specific and stays: the tool surface, outcome classification for
  the agent, the pydantic response models, the cell wiring, and a test double
  for the MCP's own unit tests. That double is a few lines returning scripted
  `BackendMotion` values, lives under `mcp/tests/`, and needs no physics: the
  service layer's tests want "stalled on close" and "timed out", not a travel
  model.

Concretely for this PR: keep `backend.py` for now with a module note that its
home is the ros client package once #61 lands, and reduce `mock_backend.py` to
a test double in `mcp/tests/`:

```python
@dataclass
class ScriptedBackend:
    name = "scripted"
    motions: list[BackendMotion]
    state: BackendState = BackendState(position_rad=0.0)

    def read_state(self) -> BackendState:
        return self.state

    def move_to(self, position_rad, max_effort_n, timeout_s) -> BackendMotion:
        return self.motions.pop(0)

    def health(self) -> BackendHealth:
        return BackendHealth(reachable=True, controller_active=True)
```

Since `GripperBackend` is a `typing.Protocol`, the double needs no import from
the ros package, and the MCP's pytest job stays ROS-free. The example wiring in
#58 then drops the `bench` mock entry, and the "no hardware" story in the README
becomes the driver on the SDK's simulated gripper (`use_dummy`, extended to
select it).

ebarnett3 (approved):

mcp/gripper_mcp/backend.py
jlongvalRobotiq: Landed on ebarnett3's version: the contract stays here for now with a note that it moves to a ros client package next to the ROS backend.

mcp/gripper_mcp/backend.py
jlongvalRobotiq: One sentence in the module docstring: `refused` is the backend rejecting the goal before any motion, `timed_out` the motion outliving its deadline.

mcp/gripper_mcp/backend.py
jlongvalRobotiq: Done as proposed. `backend.py` stays with a note that its home is a ros client package once #61 lands; `mock_backend.py` is now `tests/fakes/gripper.py`, a test double that never ships, and `backend: mock` / `object_width_mm` / the bench entry are gone from #58. The tactile mock in #62 followed it, so the demo in #69 runs on the driver alone and says so. The follow-ups you named (expose `use_dummy` as a launch arg, `makeSimulatedGripper()` in the SDK) stay open.

mcp/tests/fakes/gripper.py
jlongvalRobotiq: Applied while moving the file: a width outside the stroke raises instead of stalling at 0 rad.

mcp/gripper_mcp/mock_backend.py
jlongvalRobotiq: Applied while moving the file: `collections.abc.Callable`.

Issue #32, opened 2026-08-15
CI: generate URDF from the xacro descriptions


ebarnett3:
The gripper descriptions are authored as xacro and expanded at launch time. Nothing in CI checks that they still expand, so a broken macro or a renamed argument is only found when someone runs the driver.

Add a CI job that generates URDF from the xacro definitions and fails on any expansion error.

## Scope

- Cover the 2F-85 and 2F-140 descriptions.
- Other models are out of scope for now, but the job should be table-driven so adding one is a one-line change.

## Worth deciding

- Whether the generated URDF is only validated (expand, then `check_urdf`) or also committed as an artifact. Committing it gives non-ROS consumers a plain URDF to read, at the cost of keeping it in sync.
- Whether to validate against each supported distro's xacro, or just one.

mbegin-robotiq:
@l-tschreiber-a11y have you started working on this ?

PR #58, opened 2026-09-03
feat(mcp): gripper datasheets and cell wiring

jlongvalRobotiq:
> **Updated 2026-09-08 after @mbegin-robotiq's review.** The datasheet no longer
> carries the command joint: its name and range are read from `robot_description`
> at runtime (the joint the finger joints `mimic`), so a prefixed cell resolves by
> itself. What stays in the datasheet is the product sheet only: stroke, rated grip
> force and timing defaults. Also from the review: the datasheets ship inside the
> package, every config model rejects unknown keys, a wiring file without the
> leading `- ` is explained, and `mcp/grippers.yaml` is gitignored. Ripple in
> #59–#61, #65, #69. The original text below is edited where it was made false.

## Change

Fourth PR of the `mcp/` series (stacked on #57). Two YAML layers, deliberately split, and the code that loads them:

1. **`gripper_mcp/datasheets/<model>.yaml`** — one datasheet per Robotiq model, holding only what the robot cannot tell us: the stroke (85 mm / 140 mm), the rated grip force from the product sheet (2F-85 20–235 N, 2F-140 10–125 N) and the default requested effort and motion timeout, with the default validated against the rated range. The command joint's name and open/closed angles are the backend's to report: the ROS backend reads them from `robot_description` (#61), the mock has a built-in joint. Shipped inside the package so an installed wheel finds them. The same on every host; supporting another model is adding a file.
2. **`grippers.yaml`** — the cell: which grippers exist, their model, backend (`ros` or `mock`) and ROS namespace. Host-specific, so only `grippers.yaml.example` ships (a left 2F-85, a right 2F-140, and a mock holding a 40 mm virtual cube); `mcp/grippers.yaml` is gitignored.
3. **`config.py`** — pydantic-validated loaders. Every model forbids unknown keys, so a misspelled field fails at startup by name; a missing datasheet field or an unknown backend fails the same way; an unknown model lists the datasheets that do exist; a wiring file written without the leading `- ` is explained; duplicate gripper names are refused, since `robot_name` is how agents address a gripper.

Not carried over from gripper-mcp: the Modbus block and `${VAR}` expansion (both existed for serial device paths the ROS backend has no use for), and the tactile block (arrives with the tactile tools). The wiring file is YAML instead of JSON so both layers read the same way.

`pyyaml>=6` joins the runtime dependencies.

## Verification

- `uv run pytest`: 49 passed. Both datasheets match the product sheet, every shipped datasheet loads from inside the package (and the built wheel lists them), the example wiring loads and covers both backends, and the failure modes (unknown model, missing field, stale `command_joint` key, misspelled key, non-list wiring file, effort outside the rated range, unknown backend, duplicate name) each fail with a message naming the culprit.

## Stack
Stacked on #57: this PR shows only its own commits.


mcp/robot_specifications/robotiq_2f_85.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+geometry:
+  max_opening_mm: 85.0
+  min_opening_mm: 0.0
+
+  # The command joint's range from the URDF (limit lower/upper). 0 is OPEN.
+  # The opening is mapped onto it linearly; the real linkage is a four-bar, so
+  # mid-stroke openings are approximate.
+  knuckle_rad_open: 0.0
+  knuckle_rad_closed: 0.8
```
ebarnett3: ### 1. `knuckle_rad_closed` uses the URDF joint limit, but the driver closes at `gripper_closed_position` — High

`mcp/robot_specifications/robotiq_2f_85.yaml:17` says 0.8 and
`robotiq_2f_140.yaml:11` says 0.7, taken from `<limit upper=...>`. The driver
scales the register with a different constant:
`grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp:58-60`
returns `closed_position * (counts - 3) / 227`, and `closed_position` is the
xacro default `gripper_closed_position` = 0.7929 (`2f_85.ros2_control.xacro:21`)
and 0.695 (`2f_140.ros2_control.xacro:21`). A fully closed gripper is therefore
reported at 0.7929 rad, which #56's `knuckle_rad_to_opening_mm` turns into
0.75 mm open on the 85 and 1.0 mm on the 140. The "exact at both ends" promise
in `units.py:13` does not hold at the closed end, and `opening_fraction` never
reaches 0. Commands are unaffected (the byte clamps).

This is the failure mode of defining a conversion twice: two repos, two
languages, two "closed" values, and nothing that checks them against each
other. The fix should remove the second definition, not synchronise it.

Recommended: for the `ros` backend, take joint name and closed position from
the `/robot_description` the controller manager already runs on. The
`<ros2_control>` block carries both, next to each other
(`2f_common.ros2_control.xacro:43`):

```xml
<joint name="${prefix}robotiq_85_left_knuckle_joint">
  <param name="gripper_closed_position">${gripper_closed_position}</param>
```

mcp/robot_specifications/robotiq_2f_85.yaml
```diff
@@ -0,0 +1,23 @@
+# Robotiq 2F-85 datasheet. One file per gripper model, the same on every host;
+# cell wiring (which grippers exist, their ROS namespace) lives in grippers.yaml.
+# Supporting another Robotiq model is adding one file here, no code.
+
+model: robotiq_2f_85
+
+# The joint robotiq_gripper_controller commands, as named by robotiq_description.
+command_joint: robotiq_85_left_knuckle_joint
```
ebarnett3: ### 2. `command_joint` is per-instance, not per-model, once a `prefix` is in play — Medium

`mcp/robot_specifications/robotiq_2f_85.yaml:8` fixes the joint name in the
datasheet, and `config.py:38` makes it a model-level field. The URDF macros take
a prefix: `robotiq_2f_85_macro.urdf.xacro:249` declares
`${prefix}robotiq_85_left_knuckle_joint`, `robotiq_2f_140.xacro:153`
`${prefix}finger_joint`. #61's ROS backend looks the joint up by exact name in
`joint_states` (`ros_backend.py:107`), so a cell that prefixes its grippers
(two of them in one description is the obvious case) gets "joint not found"
with no way to fix it in `grippers.yaml`.

Fix: keep the datasheet value as the default and let the wiring override it,
mirroring how `robotiq_description` composes the name:

```python
class GripperConfig(BaseModel):
    name: str
    model: str
    backend: Backend = "mock"
    namespace: str = ""
    prefix: str = ""          # URDF joint prefix; joint = prefix + datasheet command_joint
    description: str = ""
    object_width_mm: float | None = None
```

and document it in `grippers.yaml.example`. Resolution (`spec.command_joint`
prefixed) belongs to whoever builds the backend in #59/#61; the schema is
this PR's.

mcp/gripper_mcp/config.py
```diff
@@ -0,0 +1,74 @@
...
+
+from pathlib import Path
+from typing import Literal
+
+import yaml
+from pydantic import BaseModel
+
+from gripper_mcp.units import GripperGeometry
+
+SPEC_DIR = Path(__file__).resolve().parent.parent / "robot_specifications"
```
ebarnett3: ### 3. `robot_specifications/` is outside the package and not in the wheel — Low

`config.py:19` reaches up to `mcp/robot_specifications`, and
`pyproject.toml:23` ships only `gripper_mcp`. The Docker image in #69 copies the
directory by hand so it works there; `uv build` / `pip install .` produce a
package whose `SPEC_DIR` does not exist. Move the datasheets into the package
and resolve them as package data:

```python
from importlib.resources import files
SPEC_DIR = Path(str(files("gripper_mcp") / "robot_specifications"))
```

with the YAML files under `mcp/gripper_mcp/robot_specifications/` (hatchling
includes non-Python files in the package by default). Then #69 can drop its
`COPY robot_specifications` line.

mcp/gripper_mcp/config.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def load_gripper_configs(path: Path) -> list[GripperConfig]:
+    if not path.exists():
+        raise FileNotFoundError(
+            f"Gripper wiring file not found: {path}. "
+            "Copy grippers.yaml.example next to it and edit."
+        )
+    entries = yaml.safe_load(path.read_text()) or []
+    configs = [GripperConfig.model_validate(entry) for entry in entries]
```
ebarnett3: ### 4. A mapping instead of a list gives an opaque error — Low

`config.py:67-68` iterates whatever `safe_load` returned; a user who writes
`grippers:` at the top level gets pydantic complaining that `"grippers"` is
not a valid dictionary. One `TypeAdapter` validates the shape and the entries
in one go and names the problem:

```python
from pydantic import TypeAdapter
GRIPPER_LIST = TypeAdapter(list[GripperConfig])
...
    configs = GRIPPER_LIST.validate_python(yaml.safe_load(path.read_text()) or [])
```

mcp/gripper_mcp/config.py
```diff
@@ -0,0 +1,74 @@
...
+
+from gripper_mcp.units import GripperGeometry
+
+SPEC_DIR = Path(__file__).resolve().parent.parent / "robot_specifications"
+
+
+class GripperConfig(BaseModel):
+    name: str
+    model: str
+    backend: Literal["ros", "mock"] = "mock"
```
ebarnett3: ### 5. `Backend` literal duplicated — Nit

`config.py:25` re-declares `Literal["ros", "mock"]`; `models.py:12` already
exports `Backend`. Import it so adding a backend is one edit.

mcp/gripper_mcp/config.py
```diff
@@ -0,0 +1,74 @@
...
+    namespace: str = ""
+    description: str = ""
+    object_width_mm: float | None = None
```
ebarnett3: ### 6. `backend: mock` and `object_width_mm` are the shipped mock's surface — depends on #57 finding 1

`config.py:25-28` and the `bench` entry in `grippers.yaml.example`. If the
shipped mock is replaced by the driver on the SDK fake, `use_dummy` , both
fields and the example entry go, and `Backend` collapses to `"ros"`. If it
stays, `object_width_mm` needs `Field(default=None, gt=0)`; the upper bound
against the stroke needs the geometry and is raised on #57

mcp/grippers.yaml.example
```diff
@@ -0,0 +1,25 @@
+# Cell wiring: which grippers the server exposes, and where each one's driver
+# lives. Copy to grippers.yaml and edit. Every entry needs a name (what agents
```
ebarnett3: ### 7. `grippers.yaml` is host-specific but not ignored — Nit

`grippers.yaml.example:2` says "copy to grippers.yaml and edit"; nothing stops
the copy from being committed. Add to `.gitignore`:

```
mcp/grippers.yaml
```

mcp/tests/test_config.py
```diff
@@ -0,0 +1,107 @@
...
+    load_model_specs,
+)
+
+MCP_DIR = Path(__file__).resolve().parent.parent
+EXAMPLE_WIRING = MCP_DIR / "grippers.yaml.example"
+
+MODELS = {
+    "robotiq_2f_85": ("robotiq_85_left_knuckle_joint", 85.0, 0.8),
+    "robotiq_2f_140": ("finger_joint", 140.0, 0.7),
+}
```
ebarnett3: ### 8. "matches the URDF" test compares against hand-typed numbers — Suggestion

`tests/test_config.py:17-20` and `:30-38`. The description package is in the
same repo, so the test can read the driver's closed position (finding 1)
instead of a second copy of the constant:

```python
XACRO = MCP_DIR.parent / "grippers/robotiq_description/urdf"
CONTROL_XACRO = {
    "robotiq_2f_85": XACRO / "2f_85.ros2_control.xacro",
    "robotiq_2f_140": XACRO / "2f_140.ros2_control.xacro",
}

def driver_closed_position(path: Path) -> float:
    match = re.search(r"gripper_closed_position:=([\d.]+)", path.read_text())
    return float(match.group(1))
```

then `assert spec.geometry.knuckle_rad_closed == driver_closed_position(CONTROL_XACRO[model])`.
A change to the xacro default now fails this job instead of drifting silently.
Keep the max-opening assertion literal (product spec, not in the URDF), and
the joint-name assertion can read `<joint name="${prefix}...">` from the same
file.

mbegin-robotiq (commented):
Reviewed the two YAML layers and the loader. The split itself is sound and the joint ranges check out against the URDFs, but the datasheets duplicate values `robotiq_description` already owns and the copies have already drifted. Main comment is on `robotiq_2f_85.yaml`; the rest are packaging and validation issues that stand on their own.

mcp/robot_specifications/robotiq_2f_85.yaml
mbegin-robotiq: **The datasheets duplicate the URDF, and the copies have already drifted.**

Three of the seven fields here are verbatim copies of values `robotiq_description` already owns: `command_joint`, `knuckle_rad_open`, `knuckle_rad_closed`. The drift is not hypothetical - `max_effort_n: 50.0` appears in both datasheets, but:

- 2F-85: `<limit ... effort="50">` - matches
- 2F-140: `<limit ... effort="1000">` (`grippers/robotiq_description/urdf/robotiq_2f_140.xacro:158`) - off by 20x
- `grippers/robotiq_description/config/robotiq_controllers.yaml:21`: `max_effort: 50.0   # TODO tune`

So the 50.0 actually traces to a controller default explicitly marked *TODO tune*, not to either URDF. The PR description's "straight from robotiq_description's URDF" holds for the joint ranges but not for effort: three sources, two already disagreeing. The unit label is also worth a second look - URDF revolute `effort` is N-m, while `max_effort_n` reads as newtons of grip force.

**Suggested direction:** read the joint name and limits from `/robot_description` at startup instead of copying them. `robot_state_publisher` publishes it fully expanded (prefixes and mimics resolved), and `urdf_parser_py` gives you `joint.limit.lower/upper/effort` directly. Two things fall out for free:

- the config is guaranteed to match *the robot the server is actually talking to*, which no static file can promise;
- the prefix problem goes away. Every `robotiq_description` macro is `${prefix}`-parameterized, so as written the shipped left/right example cannot address `left_robotiq_85_left_knuckle_joint` - but the expanded description already carries the real name, so no `joint_prefix` field is needed.

That shrinks a datasheet to only what the URDF genuinely lacks:

```yaml
model: robotiq_2f_140
geometry:
  max_opening_mm: 140.0    # product datasheet; the linear mm<->rad map
  min_opening_mm: 0.0      # is the four-bar approximation
defaults:
  motion_timeout_s: 10.0
```

which is a genuinely non-derivable calibration file. The mock backend has no `/robot_description`, so it will need either a supplied URDF or a small fallback - worth deciding here, since it shapes the schema.

mcp/gripper_mcp/config.py
mbegin-robotiq: `SPEC_DIR` resolves to a sibling of the package directory, but `pyproject.toml` has `[tool.hatch.build.targets.wheel] packages = ["gripper_mcp"]` - so the wheel ships the package and not the datasheets. An installed deployment finds nothing and reports `Available models: (none)`. The tests pass only because they run from the source tree, where the sibling directory exists.

Moving `robot_specifications/` inside `gripper_mcp/` and reaching it via `importlib.resources` fixes both. This still applies after the URDF change above, since the trimmed datasheets also need to ship.

mcp/gripper_mcp/config.py
```diff
@@ -0,0 +1,74 @@
...
+
+import yaml
+from pydantic import BaseModel
+
+from gripper_mcp.units import GripperGeometry
+
+SPEC_DIR = Path(__file__).resolve().parent.parent / "robot_specifications"
+
+
+class GripperConfig(BaseModel):
```
mbegin-robotiq: None of the four models set `extra="forbid"`, so pydantic silently drops unknown keys. A typo'd `namesapce: /left` yields an empty namespace and the server quietly talks to the wrong ROS target - which contradicts the PR description's "fails at startup with the field named".

Given these files are hand-edited per host, a shared base with `model_config = ConfigDict(extra="forbid")` seems worth it.

mcp/gripper_mcp/config.py
```diff
@@ -0,0 +1,74 @@
...
+    return specs
+
+
+def load_gripper_configs(path: Path) -> list[GripperConfig]:
+    if not path.exists():
+        raise FileNotFoundError(
+            f"Gripper wiring file not found: {path}. "
+            "Copy grippers.yaml.example next to it and edit."
+        )
+    entries = yaml.safe_load(path.read_text()) or []
```
mbegin-robotiq, line 110: The `yaml.safe_load` result is iterated without checking that it is a list. A single-gripper file written without the leading `- ` parses as a dict, iterates its keys, and fails with `Input should be a valid dictionary` for the string `name` - which points nowhere useful.

An `isinstance(entries, list)` check with a message naming the file would make the first-run experience a lot better.

mcp/gripper_mcp/datasheets/robotiq_2f_140.yaml
```diff
@@ -0,0 +1,15 @@
...
+command_joint: finger_joint
+
+geometry:
+  max_opening_mm: 140.0
+  min_opening_mm: 0.0
+  knuckle_rad_open: 0.0
+  knuckle_rad_closed: 0.7
+
+defaults:
+  max_effort_n: 50.0
```
mbegin-robotiq, line 14: `max_effort_n: 50.0` does not match this model's URDF (`effort="1000"` in `robotiq_2f_140.xacro:158`); it matches the 2F-85's value. See the main comment on `robotiq_2f_85.yaml` - if effort comes from the description at runtime this resolves itself, but flagging the specific value in case the datasheets stay as they are.

mcp/grippers.yaml.example
```diff
@@ -0,0 +1,25 @@
...
+# pass as robot_name) and a model (a file in robot_specifications/).
```
mbegin-robotiq: This tells users to copy the file to `grippers.yaml`, but that path is not gitignored - a host-specific wiring file will get committed by accident sooner or later. Worth adding `mcp/grippers.yaml` to `.gitignore` in this PR.

mcp/robot_specifications/robotiq_2f_85.yaml
jlongvalRobotiq: Done, with two adjustments. The command joint is now read from `robot_description` at first use: the backend picks the joint the finger joints `mimic`, which is prefix-proof, so a `left_` cell resolves by itself (there was no `joint_prefix` field to drop, the datasheet joint was simply unprefixed). Effort stays in the datasheet on purpose: the URDF `effort` is a joint torque limit in N·m (85: 50, 140: 1000), not a grip force, so reading it would raise the 2F-140 default 20x and mix units. The datasheets now carry the product-sheet rated grip force (85: 20–235 N, 140: 10–125 N) and the default is validated against it. What is left is stroke, rated force and timing, as you sketched. Those two product facts are the same ones robotiq/grippers#32's `DeviceProfile` carries; the copy here is deliberate and goes away once the driver exposes mm and N through that profile, at which point `units.py` goes with it. The mock keeps a built-in joint. The ripple went through #59–#61, #65 and #69; verified live against the launch file with mock hardware.

mcp/gripper_mcp/config.py
jlongvalRobotiq: Moved into `gripper_mcp/datasheets/` and resolved with `importlib.resources`; a test asserts the directory sits inside the package and the built wheel lists both files.

mcp/gripper_mcp/config.py
jlongvalRobotiq: Done, a `StrictModel` base with `extra="forbid"` under every config model, including the tactile block further up the stack. `namespcae` now fails by name.

mcp/gripper_mcp/config.py
jlongvalRobotiq: Done, the dict case now says the file must be a list of entries each starting with `- `, naming the file.

mcp/gripper_mcp/datasheets/robotiq_2f_140.yaml
jlongvalRobotiq: Resolved by the main thread: effort no longer traces to the URDF or the controller default. Both models keep 50 N as the requested default, now inside their own rated range.

mcp/grippers.yaml.example
jlongvalRobotiq: Added `mcp/grippers.yaml` to `.gitignore`.

mcp/gripper_mcp/config.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class StrictModel(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+
+class GripperConfig(StrictModel):
+    name: str
+    model: str
+    backend: Literal["ros", "mock"] = "mock"
```
ebarnett3: ### 2. `Backend` literal duplicated — Nit

`config.py:36` re-declares `Literal["ros", "mock"]`; `models.py:13` already
exports `Backend`. Import it so adding a backend is one edit.

mcp/gripper_mcp/units.py
```diff
@@ -8,6 +8,11 @@
   gripper's command joint. 0.0 = open, the joint's upper limit (0.8 on a 2F-85)
```
ebarnett3: ### 1. The closed angle is read from the wrong element of the robot description — High

The datasheet no longer carries the angle (good), but the replacement in #61
reads it from `<limit upper>` (`robot_description.py:6` and `:37`), and this
PR's `units.py:8` and `:17` still state "the joint's upper limit (0.8 on a
2F-85) = closed" and "both ends are exact". The driver does not close at the
joint limit. It scales the register with `gripper_closed_position`
(`robotiq_driver/include/robotiq_driver/gripper_scaling.hpp:58-60`), whose xacro
defaults are 0.7929 on the 2F-85 (`2f_85.ros2_control.xacro:21`) and 0.695 on the
2F-140 (`2f_140.ros2_control.xacro:21`), against limits of 0.8 and 0.7. A fully
closed gripper is therefore reported at 0.7929 rad, which
`knuckle_rad_to_opening_mm` turns into 0.75 mm open on the 85 and 1.0 mm on the
140, and `opening_fraction` never reaches 0. Commands are unaffected (the byte
clamps). mbegin's thread asked for the value to come from the description,
which it now does; this is about which value.

The right number is in the same document, as a hardware parameter of the
`<ros2_control>` block the controller manager runs on (expanded 2F-85 with
`use_fake_hardware:=false`):

```xml
<ros2_control name="RobotiqGripperHardwareInterface" type="system">
  <hardware>
    <plugin>robotiq_driver/RobotiqGripperHardwareInterface</plugin>
    <param name="gripper_closed_position">0.7929</param>
```

Fix, in #61's `robot_description.py`: after resolving the command joint, find
the `<ros2_control>` block that lists it and prefer that block's
`gripper_closed_position` over `limit/@upper`; fall back to the limit only when
the param is absent (`use_fake_hardware`, `sim_isaac`, where the generic mock
mirrors commands and the limit is as good as anything). Checked against the
expanded 2F-85 description: returns 0.7929 for the real-hardware expansion and
0.8 for the fake-hardware one.

```python
def closed_angle(robot: ElementTree.Element, joint: ElementTree.Element) -> float:
    name = joint.attrib["name"]
    for block in robot.iter("ros2_control"):
        if block.find(f"joint[@name='{name}']") is None:
            continue
        param = block.find("hardware/param[@name='gripper_closed_position']")
        if param is not None and param.text:
            return float(param.text)
    return float(joint.find("limit").attrib["upper"])
```

In this PR: reword `units.py:8` to "the driver's closed position
(`gripper_closed_position` in the description) = closed" and drop "both ends
are exact" at `units.py:17` unless the closed end is taken from that param. A
test in #61 that feeds the expanded 2F-85 xacro and asserts
`rad_closed == 0.7929` would have caught this.

End state unchanged: robotiq/grippers#16 / #32 put counts-to-metres in the SDK
and the description speaks metres; then none of this map exists (see #56
finding 1).

mcp/gripper_mcp/config.py
```diff
@@ -0,0 +1,120 @@
...
+    namespace: str = ""
+    description: str = ""
+    object_width_mm: float | None = None
```
ebarnett3: ### 3. `backend: mock` and `object_width_mm` are the shipped mock's surface — depends on #57 finding 1

`config.py:36-39` and the `bench` entry in `grippers.yaml.example:21-25`. If the
shipped mock is replaced by the driver on the SDK's simulated gripper (#57
finding 1), both fields and the example entry go, and `Backend` collapses to
`"ros"`. If it stays, `object_width_mm` needs `Field(default=None, gt=0)`; the
upper bound against the stroke needs the geometry and is raised on #57
(finding 2 there).

ebarnett3 (approved):

mcp/gripper_mcp/config.py
jlongvalRobotiq: The `backend` field left the config entirely (only the driver remains), so there is no literal to import. `Backend` in `models.py` keeps `"mock"` because the response models still carry the test double's name in the unit tests.

mcp/gripper_mcp/units.py
jlongvalRobotiq: Fixed in #61: `robot_description.command_joint` now prefers the `gripper_closed_position` param of the `ros2_control` block that lists the joint, and falls back to the limit only when the block has none (fake hardware, topic-based sim). Tests feed both expansions: 0.7929 real, 0.8 fake. `units.py` reworded here.

mcp/gripper_mcp/config.py
jlongvalRobotiq: Both fields and the bench entry are gone; `namespace` is now required on every entry.

PR #59, opened 2026-09-03
feat(mcp): service layer and outcome classification

jlongvalRobotiq:
## Change

Fifth PR of the `mcp/` series (stacked on #58). The layer the MCP tools will call, testable without a server or a ROS graph:

1. **`service.py`** — `GripperService`: per-gripper dispatch by name, millimetre-to-joint-angle conversion both ways through the model's datasheet stroke and the backend's joint, and the outcome classification. `list_grippers`, `get_state`, `open_fully`, `close_fully`, `grasp`, `move_to_opening`, `get_health`. Reported values round to a named precision (`MM_DECIMALS`, `RAD_DECIMALS`, `FRACTION_DECIMALS`).
2. **Outcome classification** — the one piece of domain judgement in the server, previously prose in an agent prompt. For a grasp: stopping before the fingers meet is `grasped` (a success, even though the controller reports `reached_goal: false, stalled: true`); fingers meeting with nothing between them is `closed_without_object`, whether the controller called that a stall or a reached goal. On a plain move a stall is `stalled_unexpectedly`. Refused and timed-out motions map to `refused` and `incomplete`. This is also where "aborted + stalled" from the sim configs (#52, `allow_stalling: false`) lands once the ROS backend (#61) translates it into `stalled: true`.
3. **Models** — `closed_without_object` added to `Outcome`.
4. **Review round 1** — an unknown gripper name raises `UnknownGripperError` from every entry point (one guard both lookups go through). The closed-stop margin is a per-model datasheet field, `stroke.closed_tolerance_mm` (1.5 mm on both shipped models), replacing a single 0.5 mm constant that left the 2F-140 with zero register counts of margin and read a real 2F-85 wearing TSF fingers (settling 0.75 mm short) as holding something. The no-hardware advice points at fake hardware, not the unreachable `use_dummy`.

5. **Classify from position** (Eric's thread) — the outcome is read from the achieved opening against the commanded one, with the datasheet margin; the controller's `stalled` flag is reported but no longer steers it, since on the real driver it is set on every goal (#29). A move that ends off its target in either direction is `stalled_unexpectedly`, which also catches an opening blocked part-way. An out-of-range request is still clamped but the `detail` now says so.

**Known limitation, next round:** the achieved position still comes from the action result, which Eric's bench (2026-09-11) showed echoes the commanded position before the fingers finish moving. The follow-up reads position and gOBJ (`object_status`) from the graph after the motion settles and drops the position inference for the grasp verdict; see mbegin's thread below.

No backend ships with this layer: the tests drive it through the scripted double under `tests/fakes/` (the Python mock moved there on #57).

## Verification

- `uv run pytest`: 87 passed (36 new): classify table over position, clamp note; unknown name from six entry points, tolerance validated inside the stroke, fingers met within the margin. The classifier is table-tested over every way a motion can end; the service tests drive it through the fake: grasp on a 40 mm cube (grasped, 40 mm achieved, reached_goal false), grasp on nothing (closed_without_object), plain close on the cube (stalled_unexpectedly), datasheet effort applied unless overridden, clamping, two grippers of different models opening to their own widths.

## Stack
Stacked on #58: this PR shows only its own 9 commits.

## Next
Once ros#70 lands (driver consuming the SDK's SI conversions, robotiq/grippers#32), the rad-to-mm map in `units.py` goes away and the service reads metres from the driver directly.

### DIFF
--- a/mcp/gripper_mcp/service.py
+++ b/mcp/gripper_mcp/service.py
@@ -1,0 +1,171 @@
+"""Translation between the MCP tool surface and the backends.
+
+Backends speak the command joint's angle; tools speak millimetres. Everything in
+between, unit conversion, outcome classification, per-gripper dispatch, lives
+here so it can be tested without a FastMCP server or a ROS graph.
+
+The one piece of domain judgement in this file is `classify`: a close that
+stalls before the fingers meet is a *successful grasp*, not a failure, and a
+close that stalls with the fingers together grasped nothing. That semantic used
+to live as prose in an agent prompt, where it can be forgotten; here it is in
+the type system.
+"""
+
+from datetime import datetime, timezone
+
+from gripper_mcp.backend import BackendMotion, GripperBackend
+from gripper_mcp.config import GripperConfig, GripperModelSpec
+from gripper_mcp.models import (
+    GripperHealth,
+    GripperInfo,
+    GripperMotionResult,
+    GripperState,
+    Outcome,
+)
+from gripper_mcp.units import (
+    GripperGeometry,
+    clamp_opening_mm,
+    knuckle_rad_to_opening_mm,
+    opening_mm_to_fraction,
+    opening_mm_to_knuckle_rad,
+)
+
+OPENING_TOLERANCE_MM = 0.5
+
+
+class UnknownGripperError(Exception):
+    pass
+
+
+class GripperService:
+    def __init__(
+        self,
+        specs: dict[str, GripperModelSpec],
+        configs: dict[str, GripperConfig],
+        backends: dict[str, GripperBackend],
+    ) -> None:
+        self._specs = specs
+        self._configs = configs
+        self._backends = backends
+
+    def list_grippers(self) -> list[GripperInfo]:
+        return [
+            GripperInfo(
+                robot_name=name,
+                model=config.model,
+                backend=self._backends[name].name,
+                description=config.description,
+            )
+            for name, config in self._configs.items()
+        ]
+
+    def get_state(self, robot_name: str) -> GripperState:
+        backend = self._backend(robot_name)
+        geometry = self._spec(robot_name).geometry
+        state = backend.read_state()
+        opening_mm = knuckle_rad_to_opening_mm(state.position_rad, geometry)
+
+        return GripperState(
+            robot_name=robot_name,
+            opening_mm=round(opening_mm, 2),
+            opening_fraction=round(opening_mm_to_fraction(opening_mm, geometry), 4),
+            knuckle_rad=round(state.position_rad, 4),
+            force_n=state.force_n,
+            backend=backend.name,
+            measured_at=datetime.now(timezone.utc),
+        )
+
+    def open_fully(self, robot_name: str) -> GripperMotionResult:
+        geometry = self._spec(robot_name).geometry
+        return self.move_to_opening(robot_name, geometry.max_opening_mm)
+
+    def close_fully(self, robot_name: str) -> GripperMotionResult:
+        geometry = self._spec(robot_name).geometry
+        return self.move_to_opening(robot_name, geometry.min_opening_mm)
+
+    def grasp(
+        self, robot_name: str, max_effort_n: float | None = None
+    ) -> GripperMotionResult:
+        geometry = self._spec(robot_name).geometry
+        return self.move_to_opening(
+            robot_name, geometry.min_opening_mm, max_effort_n, is_grasp=True
+        )
+
+    def move_to_opening(
+        self,
+        robot_name: str,
+        opening_mm: float,
+        max_effort_n: float | None = None,
+        is_grasp: bool = False,
+    ) -> GripperMotionResult:
+        backend = self._backend(robot_name)
+        spec = self._spec(robot_name)
+        target_mm = clamp_opening_mm(opening_mm, spec.geometry)
+        motion = backend.move_to(
+            position_rad=opening_mm_to_knuckle_rad(target_mm, spec.geometry),
+            max_effort_n=(
+                spec.defaults.max_effort_n if max_effort_n is None else max_effort_n
+            ),
+            timeout_s=spec.defaults.motion_timeout_s,
+        )
+        achieved_mm = knuckle_rad_to_opening_mm(
+            motion.final_position_rad, spec.geometry
+        )
+        stopped_on_object = stopped_on_something(motion, achieved_mm, spec.geometry)
+
+        return GripperMotionResult(
+            robot_name=robot_name,
+            commanded_opening_mm=round(target_mm, 2),
+            achieved_opening_mm=None if motion.refused else round(achieved_mm, 2),
+            reached_goal=motion.reached_goal,
+            stalled=motion.stalled,
+            object_grasped=stopped_on_object if is_grasp else None,
+            outcome=classify(motion, stopped_on_object, is_grasp),
+            detail=motion.detail,
+            backend=backend.name,
+        )
+
+    def get_health(self, robot_name: str) -> GripperHealth:
+        backend = self._backend(robot_name)
+        health = backend.health()
+
+        return GripperHealth(
+            robot_name=robot_name,
+            reachable=health.reachable,
+            controller_active=health.controller_active,
+            detail=health.detail,
+            backend=backend.name,
+        )
+
+    def _backend(self, robot_name: str) -> GripperBackend:
+        if robot_name not in self._backends:
+            available = ", ".join(self._backends) or "(none)"
+            raise UnknownGripperError(
+                f"Unknown gripper '{robot_name}'. Available: {available}"
+            )
+        return self._backends[robot_name]
+
+    def _spec(self, robot_name: str) -> GripperModelSpec:
+        return self._specs[self._configs[robot_name].model]
+
+
+def stopped_on_something(
+    motion: BackendMotion, achieved_mm: float, geometry: GripperGeometry
+) -> bool:
+    return (
+        motion.stalled and achieved_mm > geometry.min_opening_mm + OPENING_TOLERANCE_MM
+    )
+
+
+def classify(motion: BackendMotion, stopped_on_object: bool, is_grasp: bool) -> Outcome:
+    if motion.refused:
+        return "refused"
+    if motion.timed_out:
+        return "incomplete"
+    if is_grasp and (motion.stalled or motion.reached_goal):
+        return "grasped" if stopped_on_object else "closed_without_object"
+    if motion.stalled:
+        return "stalled_unexpectedly"
+    if motion.reached_goal:
+        return "reached"
+    return "incomplete"
--- a/mcp/gripper_mcp/service.py
+++ b/mcp/gripper_mcp/service.py
@@ -143,15 +143,19 @@ class GripperService:
             backend=backend.name,
         )
 
-    def _backend(self, robot_name: str) -> GripperBackend:
-        if robot_name not in self._backends:
-            available = ", ".join(self._backends) or "(none)"
+    def assert_known(self, robot_name: str) -> None:
+        if robot_name not in self._configs:
+            available = ", ".join(self._configs) or "(none)"
             raise UnknownGripperError(
                 f"Unknown gripper '{robot_name}'. Available: {available}"
             )
+
+    def _backend(self, robot_name: str) -> GripperBackend:
+        self.assert_known(robot_name)
         return self._backends[robot_name]
 
     def _spec(self, robot_name: str) -> GripperModelSpec:
+        self.assert_known(robot_name)
         return self._specs[self._configs[robot_name].model]
 
     def _geometry(self, robot_name: str):
--- a/mcp/gripper_mcp/service.py
+++ b/mcp/gripper_mcp/service.py
@@ -61,7 +61,7 @@ class GripperService:
 
     def get_state(self, robot_name: str) -> GripperState:
         backend = self._backend(robot_name)
-        geometry = self._spec(robot_name).geometry
+        geometry = self._geometry(robot_name)
         state = backend.read_state()
         opening_mm = knuckle_rad_to_opening_mm(state.position_rad, geometry)
 
@@ -76,19 +76,19 @@ class GripperService:
         )
 
     def open_fully(self, robot_name: str) -> GripperMotionResult:
-        geometry = self._spec(robot_name).geometry
-        return self.move_to_opening(robot_name, geometry.max_opening_mm)
+        stroke = self._spec(robot_name).stroke
+        return self.move_to_opening(robot_name, stroke.max_opening_mm)
 
     def close_fully(self, robot_name: str) -> GripperMotionResult:
-        geometry = self._spec(robot_name).geometry
-        return self.move_to_opening(robot_name, geometry.min_opening_mm)
+        stroke = self._spec(robot_name).stroke
+        return self.move_to_opening(robot_name, stroke.min_opening_mm)
 
     def grasp(
         self, robot_name: str, max_effort_n: float | None = None
     ) -> GripperMotionResult:
-        geometry = self._spec(robot_name).geometry
+        stroke = self._spec(robot_name).stroke
         return self.move_to_opening(
-            robot_name, geometry.min_opening_mm, max_effort_n, is_grasp=True
+            robot_name, stroke.min_opening_mm, max_effort_n, is_grasp=True
         )
 
     def move_to_opening(
@@ -100,18 +100,17 @@ class GripperService:
     ) -> GripperMotionResult:
         backend = self._backend(robot_name)
         spec = self._spec(robot_name)
-        target_mm = clamp_opening_mm(opening_mm, spec.geometry)
+        geometry = self._geometry(robot_name)
+        target_mm = clamp_opening_mm(opening_mm, geometry)
         motion = backend.move_to(
-            position_rad=opening_mm_to_knuckle_rad(target_mm, spec.geometry),
+            position_rad=opening_mm_to_knuckle_rad(target_mm, geometry),
             max_effort_n=(
                 spec.defaults.max_effort_n if max_effort_n is None else max_effort_n
             ),
             timeout_s=spec.defaults.motion_timeout_s,
         )
-        achieved_mm = knuckle_rad_to_opening_mm(
-            motion.final_position_rad, spec.geometry
-        )
-        stopped_on_object = stopped_on_something(motion, achieved_mm, spec.geometry)
+        achieved_mm = knuckle_rad_to_opening_mm(motion.final_position_rad, geometry)
+        stopped_on_object = stopped_on_something(motion, achieved_mm, geometry)
 
         return GripperMotionResult(
             robot_name=robot_name,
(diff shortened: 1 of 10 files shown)


mcp/gripper_mcp/service.py
```diff
@@ -0,0 +1,171 @@
...
+        )
+        achieved_mm = knuckle_rad_to_opening_mm(
+            motion.final_position_rad, spec.geometry
+        )
+        stopped_on_object = stopped_on_something(motion, achieved_mm, spec.geometry)
+
+        return GripperMotionResult(
+            robot_name=robot_name,
+            commanded_opening_mm=round(target_mm, 2),
+            achieved_opening_mm=None if motion.refused else round(achieved_mm, 2),
```
jlongvalRobotiq: magic numbers here.

jlongvalRobotiq: Named in 142560e: `MM_DECIMALS = 2`, `RAD_DECIMALS = 4`, `FRACTION_DECIMALS = 4` next to `OPENING_TOLERANCE_MM`, used by every `round()` in the module (get_state and the motion result).

mbegin-robotiq:
## `OPENING_TOLERANCE_MM` is standing in for a signal the gripper already reports

`stopped_on_something` decides `grasped` vs `closed_without_object` by thresholding a linearised millimetre estimate at 0.5 mm. Two problems with that constant.

The driver anchors the position register on counts 3–230, so one count is `stroke_mm / 227`:

| Model | mm/count | Margin before an empty close reads `grasped` |
|---|---|---|
| 2F-85 | 0.374 | 1 count |
| 2F-140 | **0.617** | **0 counts** |

On a 2F-140 a close landing one count short of the stop reports `grasped` on an empty gripper — and `gripper_scaling.hpp` warns that this exact off-by-one is expected ("Truncates rather than rounds, so a position read off the gripper and commanded straight back can land one count below where it came from"). The result is independent of `gripper_closed_position`, so it isn't fixable in the description.

Separately, `units.py` says to trust the mm value "to a couple of millimetres, not better" — the threshold is 4× tighter than the accuracy of the quantity it thresholds.

## But the real point is that we shouldn't be inferring this at all

The gripper reports object detection directly in gOBJ, and this repo's driver already exports it verbatim as an `object_status` state interface (`hardware_interface.cpp:286` — "gCU in amperes, and gOBJ verbatim"), reaching the graph on `/dynamic_joint_states` via `joint_state_broadcaster`. Nothing consumes it.

Meanwhile `stalled` isn't the gripper's verdict either — it's the controller's 50 ms velocity heuristic (`stall_timeout: 0.05` in `robotiq_controllers.yaml`). So we are stacking two inferences to reconstruct something the hardware states outright.

Whenever the hardware can tell us, we should use what it says.

## Suggested change here

Add a nullable object-detection field to `BackendMotion` now, plus a source marker, and have `classify` prefer it:

```python
object_detected: bool | None = None      # gOBJ, when the backend can read it
detection_source: Literal["gripper", "inferred"] = "inferred"
```

Unused until the ROS backend subscribes, but `BackendMotion`'s shape is the expensive thing to change late — it is consumed by `classify`, `ros_messages.py`, the fakes, and every test.

## On simulation

Simulation cannot produce a real gOBJ: the SDK fake has no current feedback loop to derive it from, and Isaac has the contact model but no path to carry a custom state interface through `TopicBasedSystem`. So sim keeps the millimetre heuristic — but it should report `detection_source: "inferred"` rather than synthesising a fake gOBJ value.

If we fake it, the inferred path is the only one our tests ever exercise while the real one looks proven. Faithfully replicating how gOBJ is determined is worth doing eventually; until then, best-effort and honest about it beats a convincing fake.

## Either way, the tolerance needs fixing for the fallback path

I would derive it from the geometry (`max(0.5, 2 * stroke_mm / 227)`) or make it a datasheet field, since one constant currently gives the two shipped models materially different behaviour.

mcp/gripper_mcp/service.py
```diff
@@ -0,0 +1,184 @@
...
+    def _backend(self, robot_name: str) -> GripperBackend:
+        if robot_name not in self._backends:
+            available = ", ".join(self._backends) or "(none)"
+            raise UnknownGripperError(
+                f"Unknown gripper '{robot_name}'. Available: {available}"
+            )
+        return self._backends[robot_name]
+
+    def _spec(self, robot_name: str) -> GripperModelSpec:
+        return self._specs[self._configs[robot_name].model]
```
mbegin-robotiq, line 167: `_spec` indexes `self._configs` unguarded, so it raises a bare `KeyError` instead of `UnknownGripperError`. `open_fully`, `close_fully` and `grasp` all call it *before* `self._backend(...)`, so they bypass the guard entirely:

```
get_state    -> UnknownGripperError: Unknown gripper 'nope'. Available: (none)
open_fully   -> KeyError: 'nope'
close_fully  -> KeyError: 'nope'
grasp        -> KeyError: 'nope'
get_health   -> UnknownGripperError: Unknown gripper 'nope'. Available: (none)
```

`move_to_opening` gets it right (backend first), which is why `gripper_set_opening` is fine. But open/close/grasp are exactly the tools an agent calls with a guessed name, and they get `KeyError: 'nope'` instead of the message listing the available grippers.

Simplest fix is a guard inside `_spec` itself, so ordering stops mattering.

Worth noting `test_an_unknown_gripper_is_named_with_the_available_ones` covers only `get_state` — one of the two methods that happen to be correct. Parametrising it over all five entry points both catches this and keeps it caught.

mcp/gripper_mcp/service.py
```diff
@@ -0,0 +1,184 @@
...
+
+def stopped_on_something(
+    motion: BackendMotion, achieved_mm: float, geometry: GripperGeometry
+) -> bool:
+    return (
+        motion.stalled and achieved_mm > geometry.min_opening_mm + OPENING_TOLERANCE_MM
+    )
+
+
+def classify(motion: BackendMotion, stopped_on_object: bool, is_grasp: bool) -> Outcome:
```
ebarnett3: ## 1. Object detection logic here simulates gripper firmware logic. It belongs in a simulated gripper object, rather than a high-level consumer. 

More specifically, gOBJ, the object-detection field of the status byte, is read by the driver
at every cycle (`grippers/robotiq_driver/src/hardware_interface.cpp:397`) and
exports as the `object_status` state interface
(`hardware_interface.cpp:288`). Nothing publishes it, so the MCP cannot read it
today, which is why this PR has to infer at all. The ideal shape should be:

- gOBJ is set by the grippers SDK in the simulated gripper object (already set for a real gripper)
- a ros broadcaster publishes it
- it is consumed by the mcp here

 Two issues to file for that, both outside this PR.

While waiting for this shape to be implemented, this problem should be fixed:

**On the real driver, `gripper_open`, `gripper_close` and `gripper_set_opening` always report `stalled_unexpectedly` — High**

`classify` (`service.py:173-184`) treats `motion.stalled` as a fact about the
fingers. It is not one on the ROS backend. Per #29, measured on a 2F-85, the
driver never computes the velocity state, so
`parallel_gripper_action_controller` trips stall detection on **every** goal and
the result always says `stalled: true`, `reached_goal: false`, even when the
gripper lands exactly where it was told. #61's backend passes both fields
straight through (`ros_backend.py:193-194`), so `service.py:180` fires on every
non-grasp move.

Reproduced with a double reporting what #29 measured, against the shipped 2F-85
datasheet and #61's joint geometry:

```
gripper_open: achieved=85.0 mm reached_goal=False stalled=True outcome='stalled_unexpectedly'
set_opening(30 mm) -> outcome='stalled_unexpectedly'
```

Three of the six tools therefore report an anomaly on every successful call. The
grasp path is fine, because `stopped_on_something` (`service.py:165-170`) also
requires the fingers to have stopped short, and that part is true:

```
empty air, fingers fully closed: achieved=0.0 mm  grasp -> outcome='closed_without_object'
40 mm cube:                      achieved=39.6 mm grasp -> outcome='grasped'
```

The tests do not catch it because the double in `tests/fakes/` reports
`stalled` only when it actually stopped on the virtual object, which is what the
driver ought to do rather than what it does.

Fix: decide from position, which is the signal the driver reports correctly, and
leave `stalled` in the response for the agent to see without letting it steer
the outcome.

```python
def missed_target(commanded_mm: float, achieved_mm: float) -> bool:
    return abs(achieved_mm - commanded_mm) > OPENING_TOLERANCE_MM


def stopped_on_something(commanded_mm: float, achieved_mm: float) -> bool:
    # Stopped short of a close, i.e. wider than commanded: something is in the way.
    return achieved_mm - commanded_mm > OPENING_TOLERANCE_MM


def classify(
    motion: BackendMotion, commanded_mm: float, achieved_mm: float, is_grasp: bool
) -> Outcome:
    if motion.refused:
        return "refused"
    if motion.timed_out:
        return "incomplete"
    if is_grasp:
        return (
            "grasped"
            if stopped_on_something(commanded_mm, achieved_mm)
            else "closed_without_object"
        )
    if missed_target(commanded_mm, achieved_mm):
        return "stalled_unexpectedly"
    return "reached"
```

At the call site (`service.py:118, 128-129`), `stopped_on_object` becomes
`stopped_on_something(target_mm, achieved_mm)` and `classify` takes the two
openings instead of the flag-derived boolean.

`missed_target` also catches the case the current code cannot see at all: an
*opening* blocked part-way, where `achieved` is narrower than commanded, is
today reported as `reached` whenever the stall flag happens to be false.

This does not need #29 fixed first, and it stays correct when #29 is fixed.
A regression test wants a double that lands where it was commanded and reports
`stalled: true` anyway, which is one line in `tests/fakes/`.

mcp/gripper_mcp/service.py
```diff
@@ -0,0 +1,184 @@
...
+        self,
+        robot_name: str,
+        opening_mm: float,
+        max_effort_n: float | None = None,
+        is_grasp: bool = False,
+    ) -> GripperMotionResult:
+        backend = self._backend(robot_name)
+        spec = self._spec(robot_name)
+        geometry = self._geometry(robot_name)
+        target_mm = clamp_opening_mm(opening_mm, geometry)
```
ebarnett3, line 117: ### 2. Clamping is silent — Low

`service.py:109` clamps the requested opening and `service.py:122` reports the
clamped value as `commanded_opening_mm`. An agent that asks for 200 mm on a
2F-85 gets a result saying it commanded 85 mm and reached it, with nothing
saying its request was out of range. `gripper_set_opening`'s description in #60
does mention clamping, which the agent may not re-read at call time.

Fix: say it in `detail`, which is otherwise only the backend's message.

```python
detail=(
    motion.detail
    if target_mm == opening_mm
    else f"Requested {opening_mm:.1f} mm, clamped to {target_mm:.1f} mm. {motion.detail}"
),
```

mcp/gripper_mcp/service.py
```diff
@@ -0,0 +1,184 @@
...
+            available = ", ".join(self._backends) or "(none)"
+            raise UnknownGripperError(
+                f"Unknown gripper '{robot_name}'. Available: {available}"
+            )
+        return self._backends[robot_name]
+
+    def _spec(self, robot_name: str) -> GripperModelSpec:
+        return self._specs[self._configs[robot_name].model]
+
+    def _geometry(self, robot_name: str):
```
ebarnett3: ### 3. `_geometry` has no return annotation — Nit

`service.py:157`. Every other method is annotated; this one returns
`GripperGeometry`, already imported.

mcp/gripper_mcp/service.py
```diff
@@ -0,0 +1,184 @@
...
+def geometry_of(spec: GripperModelSpec, backend: GripperBackend) -> GripperGeometry:
+    return GripperGeometry.of(spec.stroke, backend.joint_geometry())
+
+
+def stopped_on_something(
+    motion: BackendMotion, achieved_mm: float, geometry: GripperGeometry
+) -> bool:
+    return (
+        motion.stalled and achieved_mm > geometry.min_opening_mm + OPENING_TOLERANCE_MM
+    )
```
ebarnett3: On the real gripper, the stalled vs. reached-position logic (object detection) is reported by the gripper firmware in the status block. So I think any simulated gripper should report this in the same place, rather than implementing object detection logic at higher levels of abstraction such as the ros and the mcp server.

mcp/gripper_mcp/service.py
```diff
@@ -0,0 +1,184 @@
...
+)
+from gripper_mcp.units import (
+    GripperGeometry,
+    clamp_opening_mm,
+    knuckle_rad_to_opening_mm,
+    opening_mm_to_fraction,
+    opening_mm_to_knuckle_rad,
+)
+
+OPENING_TOLERANCE_MM = 0.5
```
ebarnett3: ### One 0.5 mm tolerance for every model is under one register count on a 2F-140 — Medium

`OPENING_TOLERANCE_MM = 0.5` (`service.py:33`) is the margin for every
position comparison, on every model. The driver's position is a byte over the
usable travel band, counts 3 to 230
(`grippers/robotiq_driver/include/robotiq_driver/gripper_scaling.hpp:47-60`), so
what the MCP sees is quantised:

| model | 1 register count | 0.5 mm is |
|---|---|---|
| 2F-85 | 0.3744 mm | 1.34 counts |
| 2F-140 | 0.6167 mm | 0.81 counts |

On a 2F-140 the tolerance is smaller than the quantum. Commanding and reading
back each round to a count, so a move that lands perfectly can still report up
to half a count of difference, 0.31 mm there, leaving under 0.2 mm of margin
before a correct move is called a miss. Add the linear mm-to-rad approximation,
which the module says is good to a couple of millimetres mid-stroke
(`units.py:16-17`), and the margin is gone.

Fix: derive the tolerance from the model rather than fixing it in millimetres
for all of them. Two counts is a defensible floor, and the stroke is already in
the datasheet:

```python
USABLE_COUNTS = 227  # register band 3..230, see gripper_scaling.hpp
TOLERANCE_COUNTS = 2


def tolerance_mm(stroke: Stroke) -> float:
    span = stroke.max_opening_mm - stroke.min_opening_mm
    return TOLERANCE_COUNTS * span / USABLE_COUNTS
```

0.75 mm on a 2F-85, 1.23 mm on a 2F-140. It is also a reason not to lean on
position at all in the long run: gOBJ is not quantised and does not need a
tolerance.

ebarnett3 (approved):

mcp/gripper_mcp/service.py
jlongvalRobotiq: Fixed. `assert_known` is now the single guard and both `_spec` and `_backend` go through it, so call order stops mattering. The unknown-name test is parametrised over all six entry points, `move_to_opening` included.

jlongvalRobotiq:
@mbegin-robotiq, re your comment on `OPENING_TOLERANCE_MM` and gOBJ (https://github.com/robotiq/ros/pull/59#issuecomment-5624184248):

Agreed on both halves, and Eric did some hardware bench tests that made the second one urgent: on the real driver `stalled` is always true (#29) and the action result echoes the commanded position, so the classifier's two inputs are both wrong on hardware. Two moves here:

1. In this round: the constant is gone. The margin is a datasheet field, `stroke.closed_tolerance_mm`, 1.5 mm on both models (Eric measured a 2F-85 with TSF fingers settling 0.75 mm short of the description's closed angle; two counts on the 2F-140 is 1.23 mm). I went with a datasheet field rather than `2 * stroke / 227` so the register size never enters the MCP.
2. Next round, as its own change: `classify` takes gOBJ as the verdict, `None` when the backend cannot supply it, no inferred fallback at all. Eric's position (and the SDK enum he quoted on #60) is that the MCP should not infer this, and I agree. Eventually, this could be a functionality from the omnipuch. That also brings an `Outcome` term for a hold achieved by opening (`DetectedWhileOpening`). #61 reads position and `object_status` from `/dynamic_joint_states` after the motion settles rather than from the action result. Tracking it on our side; I'll link the PR here.

mcp/gripper_mcp/service.py
jlongvalRobotiq: Done, along the lines you sketched. `classify` now takes the commanded and achieved openings and the model's `Stroke`; `missed_target` and `stopped_on_something` compare them against the datasheet margin, and `stalled` is reported but does not steer the outcome. The fake has an `always_stalled` switch that reports what #29 measured, and the service tests run open, grasp-on-air and grasp-on-cube through it. The part-way-blocked opening is covered too.

Agreed on the ideal shape (SDK sim sets gOBJ, a broadcaster publishes it, the MCP consumes it) and that it is two issues outside this PR; both are on our list. One caveat your bench surfaced that this fix does not cover: `achieved_mm` still comes from the action result, which echoes the commanded position ~100 ms in, so on hardware a move reads as `reached` regardless. Reading position and `object_status` from the graph after settle is the next change, in #61.

mcp/gripper_mcp/service.py
jlongvalRobotiq: Done: `detail` now starts with "Requested 200.0 mm, clamped to 85.0 mm. " when the request was out of range, and is the backend's message verbatim otherwise.

mcp/gripper_mcp/service.py
jlongvalRobotiq: The wrapper is gone rather than annotated: `geometry_of(spec, backend)` is called at its two sites. (Annotating it tipped the class over the quality gate's coupling limit, and the one-liner was not earning its place.)

mcp/gripper_mcp/service.py
jlongvalRobotiq: Agreed. The MCP should consume gOBJ, not reconstruct it, and the simulated gripper that sets it belongs in the SDK so every consumer sees the same behaviour. The position comparison that remains in `classify` after this round is an explicit stopgap and the module docstring says so; it goes when a broadcaster carries `object_status`. Filing the SDK and broadcaster issues on our side.

mcp/gripper_mcp/service.py
jlongvalRobotiq: The constant is gone. The margin is a datasheet field per model, `stroke.closed_tolerance_mm`, used for every position comparison. I set 1.5 mm on both models rather than exactly two counts (0.75 / 1.23 mm): on the 2F-85 your own measurement had TSF fingers settling 0.75 mm short, right at two counts, so it needed headroom, and nobody has measured the 2F-140's settle. Datasheet field rather than a `stroke / 227` formula so the register size stays out of the MCP. Agreed that gOBJ makes the whole question go away.

mcp/tests/fakes/gripper.py
```diff
@@ -7,9 +7,14 @@
...
 
+`always_stalled` makes it report what the real driver reports today
+(robotiq/ros#29): `stalled: true, reached_goal: false` on every goal, wherever
+the fingers actually ended up.
+
 It is deliberately not part of the shipped package. Running the MCP without
-hardware is the ROS driver's job, on the SDK's fake gripper (`use_dummy`), so
-that the driver, the controller and the action are exercised too.
+hardware is the ROS driver's job, on ros2_control's fake hardware
+(`use_fake_hardware`, see mcp/demo), so that the controller and the action are
```
jlongvalRobotiq: mcp/demo does not yet exist.

mcp/gripper_mcp/service.py
jlongvalRobotiq: Follow-up: I dropped `always_stalled` from the fake again. Modelling a driver bug (#29) in the test double felt like the wrong layer to carry it; the classify table keeps a "landed but the driver says stalled" row, which is the guarantee that matters. Tip is now 66c5c5a.

mcp/tests/fakes/gripper.py
jlongvalRobotiq: Dropped here; the pointer comes back in #69 with the demo itself.

PR #60, opened 2026-09-03
feat(mcp): MCP tool surface on FastMCP

jlongvalRobotiq:
## Change

Sixth PR of the `mcp/` series (stacked on #59, now merged). The server itself, in three commits: the argument rename, the tool surface, then the review-driven fold of `gripper_grasp` into the motion tools.

1. **`gripper_name`** — every tool and every response model addresses a gripper by `gripper_name` (was `robot_name`). The server is independent of any particular agent stack, so the argument is named for what it is; an agent layer that wants one addressing word across servers can map it there.
2. **`server.py`** — six tools on FastMCP: `gripper_list_grippers`, `gripper_get_state`, `gripper_open`, `gripper_close`, `gripper_move_to`, `gripper_get_health`. Every tool takes an explicit `gripper_name`; nothing fans out. The `gripper_` prefix stays because MCP tool lists are flat (no namespace on the wire) and these tools sit next to other servers' in an agent's list. Tools that reduce the opening are annotated destructive, reads are read-only. The descriptions point at the `outcome` field rather than restating how it is computed. `gripper_list_grippers` reports each model's `max_opening_mm`, so `gripper_move_to` is usable without a probing round-trip.
3. **No work at import time** — `build_service(wiring, make_backend=...)` and `build_mcp(service)` assemble the server; `main()` parses arguments and runs it. The backend factory is a parameter: the default refuses every wiring entry by name until the ROS backend lands (next PR), and the tests hand in the `tests/fakes/` double and exercise one call end to end through fastmcp's in-memory client, no port involved.
4. **`uv run gripper-mcp --config grippers.yaml`** serves streamable-http on `127.0.0.1:8300`; `--host 0.0.0.0` is the deliberate flag for the cell network, and the README says the tools carry no authentication. README rewritten to describe the tool surface, the outcome vocabulary, the quick start, and how to connect Claude Code (`claude mcp add`) and Codex (`config.toml`).
5. **One close, neutral outcomes** (Eric's review) — `gripper_grasp` is gone. A grasp can happen in either direction, so the service no longer decides whether a stop was wanted: every motion reports `stopped_on_object` with `object_detected: true` when the fingers halted short of the commanded position, `reached` otherwise, mirroring the SDK's `ObjectDetection` enum. `max_effort_n` is an optional argument on `gripper_open`, `gripper_close` and `gripper_move_to`. Whether contact was a grasp or an obstruction is the caller's judgement.

Not carried over from gripper-mcp: `gripper_activate`, `gripper_recover_from_fault` (the ROS driver activates on bring-up and exposes no fault register), `gripper_get_tip_geometry` (dropped in #59), and the tactile tools (later PRs). Speed and force percentages are gone; the motion tools take an optional `max_effort_n`, which is what the `gripper_cmd` action actually accepts.

`fastmcp` joins the runtime dependencies (4.0.2).

## Verification

- `uv run pytest`: 93 passed. Registered tool set pinned, every tool but the listing takes `gripper_name`, every description is non-empty and clean of indentation, destructive/read-only annotations checked, a close called through the MCP client on a fake holding a 40 mm cube comes back `stopped_on_object` at 40 mm with `object_detected: true`, function-scoped fixtures so no test leaves state for the next, and the default factory names the missing ROS backend.
- `uv run gripper-mcp --config grippers.yaml --port 8399`: starts and serves `http://127.0.0.1:8399/mcp`.

## Stack
#59 is merged; this PR shows its own 3 commits.

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -1,14 +1,87 @@
 # gripper_mcp
 
 MCP server for Robotiq 2F adaptive grippers. It gives an LLM agent a small,
-gripper-shaped tool surface (open, close, grasp, read state) and drives the
-`ros2_control` driver in [`grippers/`](../grippers/) underneath, so the same
-tools work against real hardware and against a simulator launched with
-`sim_isaac:=true`.
+gripper-shaped tool surface and drives the `ros2_control` driver in
+[`grippers/`](../grippers/) underneath, so the same tools work against real
+hardware and against a simulator launched with `sim_topic_based:=true`.
 
-Work in progress: this directory currently holds the project skeleton only. The
-tool surface, the ROS 2 backend and the tactile tools land in follow-up pull
-requests.
+## Tools
+
+| Tool | What it does |
+|---|---|
+| `gripper_list_grippers` | Discover configured gripper names |
+| `gripper_get_state` | Opening (mm, fraction, joint rad) and grip force |
+| `gripper_open` | Open fully |
+| `gripper_close` | Close fully; a stall is an unexpected obstruction |
+| `gripper_move_to` | Move to a specific position, in mm of opening |
+| `gripper_grasp` | Close onto an object; stopping before the fingers meet is a success |
+| `gripper_get_health` | Reachable, controller active |
+
+Every tool takes an explicit `gripper_name`; nothing fans out to every gripper.
+Openings are in **millimetres**: `0.0` closed, the model's max opening (`85.0`
+on a 2F-85, `140.0` on a 2F-140) fully open. Every motion result carries an
+`outcome`:
+
+| Outcome | Meaning |
+|---|---|
+| `reached` | Fingers arrived at the commanded opening |
+| `grasped` | A grasp stopped on something before the fingers met (`reached_goal: false`, `stalled: true`; this is success) |
+| `closed_without_object` | A grasp closed fully; nothing was between the fingers |
+| `stalled_unexpectedly` | A plain move stopped early against resistance |
+| `incomplete` | Timed out or never finished |
+| `refused` | The backend rejected the goal |
+
+## Quick start
+
+```bash
+cd mcp
+cp grippers.yaml.example grippers.yaml   # one entry per gripper, with its namespace
+uv run gripper-mcp --config grippers.yaml
+```
+
+The server is then at `http://127.0.0.1:8300/mcp` (streamable-http); point any
+MCP client at it. The tools carry no authentication: anything that can reach
+the port can close a gripper. The default binds localhost only; pass
+`--host 0.0.0.0` to serve the cell network, and firewall the port when you do.
+The tool surface is complete in this build; the backend that
+talks to `robotiq_gripper_controller` lands in the next one, so until then
+`build_service` refuses every wiring entry by name. Without hardware, run the
+driver on ros2_control's fake hardware (`use_fake_hardware`) rather than a
+Python stand-in: that exercises the driver, the controller and the action,
+which a mock never will. The unit tests use a scripted double under
+`tests/fakes/`.
+
+### Connecting an agent
+
+[Claude Code](https://code.claude.com/docs/en/mcp), once, then start a new
+session:
+
+```bash
+claude mcp add --transport http robotiq-gripper http://127.0.0.1:8300/mcp
+```
+
+[Codex](https://developers.openai.com/codex/mcp), in `~/.codex/config.toml`
+(or `.codex/config.toml` in the project):
+
+```toml
+[mcp_servers.robotiq-gripper]
+url = "http://127.0.0.1:8300/mcp"
+```
+
+Either client then lists the `gripper_*` tools; `/mcp` in the session shows
+the connection. Ask in plain language ("open the left gripper to 30 mm") and
+the agent picks the tool.
+
+## Configuration
+
+Two layers, deliberately split:
+
+- `gripper_mcp/datasheets/<model>.yaml`: one datasheet per Robotiq model, the same
+  on every host. The command joint's name and range come straight from
+  `robotiq_description`'s URDF. Supporting another model is adding a file.
+- `grippers.yaml`: the cell's wiring, which grippers exist, their model and the
+  ROS namespace their driver runs under. Host-specific, so only
+  `grippers.yaml.example` ships.
 
 ## Not a ROS package
 
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -11,10 +11,9 @@ hardware and against a simulator launched with `sim_topic_based:=true`.
 |---|---|
 | `gripper_list_grippers` | Discover configured gripper names |
 | `gripper_get_state` | Opening (mm, fraction, joint rad) and grip force |
-| `gripper_open` | Open fully |
-| `gripper_close` | Close fully; a stall is an unexpected obstruction |
-| `gripper_move_to` | Move to a specific position, in mm of opening |
-| `gripper_grasp` | Close onto an object; stopping before the fingers meet is a success |
+| `gripper_open` | Open fully; optional `max_effort_n` |
+| `gripper_close` | Close fully; optional `max_effort_n` |
+| `gripper_move_to` | Move to a specific position, in mm of opening; optional `max_effort_n` |
 | `gripper_get_health` | Reachable, controller active |
 
 Every tool takes an explicit `gripper_name`; nothing fans out to every gripper.
@@ -25,9 +24,7 @@ on a 2F-85, `140.0` on a 2F-140) fully open. Every motion result carries an
 | Outcome | Meaning |
 |---|---|
 | `reached` | Fingers arrived at the commanded opening |
-| `grasped` | A grasp stopped on something before the fingers met (`reached_goal: false`, `stalled: true`; this is success) |
-| `closed_without_object` | A grasp closed fully; nothing was between the fingers |
-| `stalled_unexpectedly` | A plain move stopped early against resistance |
+| `stopped_on_object` | The fingers stopped on something before the commanded position, either direction; `object_detected: true`. Whether that is a grasp is the caller's call |
 | `incomplete` | Timed out or never finished |
 | `refused` | The backend rejected the goal |
 
--- a/mcp/gripper_mcp/server.py
+++ b/mcp/gripper_mcp/server.py
@@ -1,0 +1,237 @@
+"""MCP server for Robotiq 2F adaptive grippers, on the ROS 2 driver in this repo.
+
+Every tool takes an explicit `gripper_name` and nothing ever fans out to every
+configured gripper. The server is assembled by `build_mcp` from a `GripperService`
+so tests can register the tools against a test double without touching argv or
+a port; `build_service` takes the backend factory for the same reason.
+"""
+
+import argparse
+from collections.abc import Callable
+from inspect import cleandoc
+from pathlib import Path
+
+from fastmcp import FastMCP
+
+from gripper_mcp.backend import GripperBackend
+from gripper_mcp.config import (
+    SPEC_DIR,
+    GripperConfig,
+    GripperModelSpec,
+    load_gripper_configs,
+    load_model_specs,
+)
+from gripper_mcp.models import (
+    GripperHealth,
+    GripperInfo,
+    GripperMotionResult,
+    GripperState,
+)
+from gripper_mcp.service import GripperService
+
+DEFAULT_WIRING = Path("grippers.yaml")
+DEFAULT_PORT = 8300
+
+INSTRUCTIONS = (
+    "Tools for controlling Robotiq 2F adaptive grippers. Call gripper_list_grippers "
+    "to discover available names, and pass one as gripper_name to every other tool. "
+    "Openings are in millimetres: 0.0 is fully closed; fully open is the model's max "
+    "opening, reported by gripper_list_grippers. Every motion returns an outcome "
+    "field that says how it ended; read that rather than inferring from the flags."
+)
+
+READ_ONLY = {
+    "read_only_hint": True,
+    "destructive_hint": False,
+    "idempotent_hint": True,
+    "open_world_hint": True,
+}
+OPENS = {
+    "read_only_hint": False,
+    "destructive_hint": False,
+    "idempotent_hint": True,
+    "open_world_hint": True,
+}
+CLOSES = {
+    "read_only_hint": False,
+    "destructive_hint": True,
+    "idempotent_hint": True,
+    "open_world_hint": True,
+}
+GRASPS = {**CLOSES, "idempotent_hint": False}
+
+
+def describe(text: str) -> str:
+    return cleandoc(text).strip()
+
+
+BackendFactory = Callable[[GripperConfig, GripperModelSpec], GripperBackend]
+
+
+def build_backend(config: GripperConfig, spec: GripperModelSpec) -> GripperBackend:
+    raise NotImplementedError(
+        f"Gripper '{config.name}' needs the ROS backend, which is not in this build."
+    )
+
+
+def build_service(
+    wiring: Path,
+    spec_dir: Path = SPEC_DIR,
+    make_backend: BackendFactory = build_backend,
+) -> GripperService:
+    configs = {cfg.name: cfg for cfg in load_gripper_configs(wiring)}
+    specs = load_model_specs(spec_dir, {cfg.model for cfg in configs.values()})
+    backends = {
+        name: make_backend(cfg, specs[cfg.model]) for name, cfg in configs.items()
+    }
+    return GripperService(specs=specs, configs=configs, backends=backends)
+
+
+def build_mcp(service: GripperService) -> FastMCP:
+    mcp = FastMCP("robotiq_gripper_mcp", instructions=INSTRUCTIONS)
+
+    @mcp.tool(
+        name="gripper_list_grippers",
+        annotations={**READ_ONLY, "open_world_hint": False},
+        description=describe(
+            """
+        List all configured grippers.
+
+        Returns each gripper's name, model, backend and fully-open width in
+        millimetres. Use the returned names as the gripper_name argument for
+        every other tool.
+            """
+        ),
+    )
+    def gripper_list_grippers() -> list[GripperInfo]:
+        return service.list_grippers()
+
+    @mcp.tool(
+        name="gripper_get_state",
+        annotations=READ_ONLY,
+        description=describe(
+            """
+        Read the current opening of one gripper.
+
+        Returns the opening in millimetres (0.0 closed, the model's max opening
+        fully open), as a fraction, and as the controller's joint angle in
+        radians. The millimetre value comes through a linear approximation of the
+        finger linkage; trust it to a couple of millimetres, not better.
+
+        Args:
+            gripper_name: Name of the gripper (see gripper_list_grippers).
+            """
+        ),
+    )
+    def gripper_get_state(gripper_name: str) -> GripperState:
+        return service.get_state(gripper_name)
+
+    @mcp.tool(
+        name="gripper_open",
+        annotations=OPENS,
+        description=describe(
+            """
+        Open one gripper fully, releasing anything it holds.
+
+        Args:
+            gripper_name: Name of the gripper (see gripper_list_grippers).
+            """
+        ),
+    )
+    def gripper_open(gripper_name: str) -> GripperMotionResult:
+        return service.open_fully(gripper_name)
+
+    @mcp.tool(
+        name="gripper_close",
+        annotations=CLOSES,
+        description=describe(
+            """
+        Close one gripper fully.
+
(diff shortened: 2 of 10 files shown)


mcp/README.md
```diff
@@ -1,14 +1,62 @@
...
+## Quick start
+
+```bash
+cd mcp
+cp grippers.yaml.example grippers.yaml   # one entry per gripper, with its namespace
+uv run gripper-mcp --config grippers.yaml
+```
+
+The server is then at `http://127.0.0.1:8300/mcp` (streamable-http); point any
+MCP client at it. The tool surface is complete in this build; the backend that
```
jlongvalRobotiq: maybe we should add examples of how to make it work with claude code and codex ?

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,238 @@
+"""MCP server for Robotiq 2F adaptive grippers, on the ROS 2 driver in this repo.
```
jlongvalRobotiq, line 1: should the we name it something like ros_server to make it clear that this server interfaces with ros only ? Or maybe its yagni until we have a server for something else like the sdk directly ?

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,238 @@
...
+        name: make_backend(cfg, specs[cfg.model]) for name, cfg in configs.items()
+    }
+    return GripperService(specs=specs, configs=configs, backends=backends)
+
+
+def build_mcp(service: GripperService) -> FastMCP:
+    mcp = FastMCP("robotiq_gripper_mcp", instructions=INSTRUCTIONS)
+
+    @mcp.tool(
+        name="gripper_list_grippers",
```
jlongvalRobotiq, line 95: do we need to prefix everything with gripper_ ?

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,238 @@
...
+            """
+        Read the current opening of one gripper.
+
+        Returns the opening in millimetres (0.0 closed, the model's max opening
+        fully open), as a fraction, and as the controller's joint angle in
+        radians. The millimetre value comes through a linear approximation of the
+        finger linkage; trust it to a couple of millimetres, not better.
+
+        Args:
+            robot_name: Name of the gripper (see gripper_list_grippers).
```
jlongvalRobotiq: im commenting here but this comment applies to other places in this PR: we say robot_name but we really mean gripper_name ? Or is this because of ROS ?

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,238 @@
...
+        Returns each gripper's name, model and backend. Use the returned names as
+        the robot_name argument for every other tool.
+            """
+        ),
+    )
+    def gripper_list_grippers() -> list[GripperInfo]:
+        return service.list_grippers()
+
+    @mcp.tool(
+        name="gripper_get_state",
```
jlongvalRobotiq, line 111: could it be get_gripper_state or do we need regularity in the naming ?

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,238 @@
...
+        Args:
+            robot_name: Name of the gripper (see gripper_list_grippers).
+            """
+        ),
+    )
+    def gripper_get_state(robot_name: str) -> GripperState:
+        return service.get_state(robot_name)
+
+    @mcp.tool(
+        name="gripper_open",
```
jlongvalRobotiq, line 131: the gripper_ here makes sense if open would be read alone by itself... if we are in the namespace of gripper then its not required i think .

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,238 @@
...
+        Args:
+            robot_name: Name of the gripper (see gripper_list_grippers).
+            """
+        ),
+    )
+    def gripper_open(robot_name: str) -> GripperMotionResult:
+        return service.open_fully(robot_name)
+
+    @mcp.tool(
+        name="gripper_close",
```
jlongvalRobotiq, line 154: same logic applies here.

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,238 @@
...
+        Args:
+            robot_name: Name of the gripper (see gripper_list_grippers).
+            """
+        ),
+    )
+    def gripper_close(robot_name: str) -> GripperMotionResult:
+        return service.close_fully(robot_name)
+
+    @mcp.tool(
+        name="gripper_set_opening",
```
jlongvalRobotiq: i think we could try and use the move_to naming ? We use it in the sdk ?

mcp/README.md
jlongvalRobotiq: Added a "Connecting an agent" section under the quick start: `claude mcp add --transport http robotiq-gripper http://127.0.0.1:8300/mcp` for Claude Code, and the `[mcp_servers.robotiq-gripper] url = ...` block in `~/.codex/config.toml` for Codex.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Kept `server.py`: since #57's review the module is the tool surface plus wiring and takes its backend factory as a parameter, so it imports nothing from ROS. A server on the SDK directly would be a new backend behind the same module, not a second server, so `ros_server` would name it wrong. YAGNI as you said.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Kept the prefix. MCP has no depth: `tools/list` is one flat list per server with no namespace or group field, and an agent typically sees several servers' tools side by side. Some clients prefix with the server name on their own (Claude Code does), most agent frameworks do not, so `gripper_` is the only grouping we control. `robotiq_gripper_` would repeat the server's own name in every tool call for little gain.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Renamed to `gripper_name` on every tool and in every response model, here and up the stack. It was `robot_name` for regularity with other MCP servers in the franka cookbook, but this server is independent of that stack; if the cookbook needs one addressing word, its agent layer is the place to map it.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Kept prefix-first (`gripper_get_state`): it keeps the tools grouped and sorted in any client's list. Follows from the thread above on the prefix.

mcp/gripper_mcp/server.py
jlongvalRobotiq: See the prefix thread: there is no gripper namespace on the wire, so `open` would stand alone next to every other server's tools.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Same as `gripper_open`, see the prefix thread.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Kept `gripper_set_opening`. The SDK has no `moveTo` in its API (only the `GoTo` register command and a `moveTo` helper in one example), so there is no name to match, and "set the opening" says the unit domain that `opening_mm` carries. The backend protocol keeps `move_to` internally.

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,240 @@
...
+        return service.get_health(gripper_name)
+
+    return mcp
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Robotiq gripper MCP server")
+    parser.add_argument("--config", type=Path, default=DEFAULT_WIRING, metavar="PATH")
+    parser.add_argument("--spec-dir", type=Path, default=SPEC_DIR, metavar="DIR")
+    parser.add_argument("--host", default="0.0.0.0")
```
mbegin-robotiq: The default binds every interface, and `FastMCP(...)` above is constructed with no auth — so any host routable to the cell network can close a gripper.

The README says the opposite:

> The server is then at `http://127.0.0.1:8300/mcp`

Binding wide may well be what a cell wants, but documenting it as localhost is how it ends up deployed unexamined. Either default to `127.0.0.1` and make `0.0.0.0` a deliberate flag, or keep this default and say plainly in the README that the tools are unauthenticated and the port needs firewalling.

Raising it here because this is the PR that introduces `main()`.

mcp/README.md
```diff
@@ -1,14 +1,83 @@
...
+talks to `robotiq_gripper_controller` lands in the next one, so until then
+`build_service` refuses every wiring entry by name. Without hardware, run the
+driver on the SDK's fake gripper (`use_dummy`) rather than a Python stand-in:
```
mbegin-robotiq: `use_dummy` cannot currently be reached. `2f_common.ros2_control.xacro` hardcodes it:

```xml
<!-- Set use_dummy to true to connect to a dummy driver for testing purposes. -->
<param name="use_dummy">false</param>
```

There is no xacro arg and no launch arg, so following this instruction means hand-editing an installed description. `grippers.yaml.example` gives the same advice.

Everything below the param is already plumbed (`hardware_parameters.cpp:191` -> `hardware_interface.cpp:146` -> `makeFakeGripper`), so exposing it is a one-liner — and it is the same one-liner needed to get `object_status` onto the graph without hardware. Worth doing rather than un-documenting.

mcp/tests/test_server.py
```diff
@@ -0,0 +1,107 @@
...
+
+def gripper_holding_a_cube(config, spec):
+    return MockGripperBackend(
+        object_width_mm=CUBE_WIDTH_MM,
+        stroke=spec.stroke,
+        sleep_fn=lambda _seconds: None,
+    )
+
+
+@pytest.fixture(scope="module")
```
mbegin-robotiq: Module-scoped fixture over a *stateful* double: `test_a_grasp_through_the_wire_stops_on_the_virtual_cube` leaves the gripper closed on the cube for every test after it.

Harmless today since nothing later reads position, but it makes test order load-bearing. Function scope, or reset the backend between tests.

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,240 @@
...
+
+        Args:
+            gripper_name: Name of the gripper (see gripper_list_grippers).
+            """
+        ),
+    )
+    def gripper_get_health(gripper_name: str) -> GripperHealth:
+        return service.get_health(gripper_name)
+
+    return mcp
```
mbegin-robotiq, line 219: Question rather than a change request: there is no stop/abort tool.

Once the ROS backend lands a motion can occupy a worker for the full motion timeout, and an agent that changes its mind — or a human watching it close on the wrong thing — has no tool to interrupt it. Is that deliberate scope-limiting because `gripper_cmd` has no cancel surface the driver honours, or just not reached yet?

mbegin-robotiq (approved):
Approving — the tool surface is well judged. Every tool taking an explicit `gripper_name` with no fan-out is the right call for a server that sits next to others in a flat tool list, the `robot_name` -> `gripper_name` rename is clean (no leftovers anywhere), and putting the grasp semantics into the tool descriptions *and* the `outcome` field means an agent has to work to get it wrong.

Verified the error surface end to end against fastmcp 4.0.2: exception messages are not masked, so the `UnknownGripperError` text reaches the client intact. That makes the guard worth having — and makes the `KeyError` on #59 more visible, since `gripper_open` currently hands the agent the bare string `'right'`.

Four comments below: the bind address is the one I would like settled before this ships, the rest are follow-ups.

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,240 @@
...
+from gripper_mcp.service import GripperService
+
+DEFAULT_WIRING = Path("grippers.yaml")
+DEFAULT_PORT = 8300
+
+INSTRUCTIONS = (
+    "Tools for controlling Robotiq 2F adaptive grippers. Call gripper_list_grippers "
+    "to discover available names, and pass one as gripper_name to every other tool. "
+    "Openings are in millimetres: 0.0 is fully closed; fully open is the model's max "
+    "opening, reported by gripper_list_grippers. On a grasp, reached_goal=false "
```
ebarnett3: ### 1. The instructions send the agent to a tool that does not report the max opening — Medium

`server.py:38-39` tells the agent: "fully open is the model's max opening,
reported by gripper_list_grippers". `GripperInfo` (`models.py:27-31`) carries
`gripper_name`, `model`, `backend` and `description`, and nothing else. There is
no opening range anywhere in the tool surface, so an agent that wants to set a
specific opening on an unknown model has to guess, or infer the maximum by
opening fully and reading the state back.

The stroke is right there in the datasheet the service already holds, and the
tool is the natural place for it.

```python
class GripperInfo(BaseModel):
    gripper_name: str
    model: str
    backend: Backend
    max_opening_mm: float = Field(description="Fully open, in millimetres")
    description: str
```

filled in `GripperService.list_grippers` from `self._spec(name).stroke`. Then
the instruction is true and `gripper_set_opening` becomes usable without a
probing round-trip.

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,240 @@
...
+    "with stalled=true means the fingers stopped on an object: that is SUCCESS, not "
+    "a failure. Read the outcome field, it already says which."
```
ebarnett3: ### 2. The grasp semantics are stated in three places, and two of them will be wrong — Medium

`server.py:39-41` ("reached_goal=false with stalled=true means the fingers
stopped on an object: that is SUCCESS") and `gripper_grasp`'s description
(`server.py:188-193`, same claim in more words) both restate what
`GripperService.classify` decides. #59 finding 1 shows the stall flag does not
mean that on the real driver, so the prose teaches the agent a rule that
misfires, and it will keep saying so after the classification is fixed.

The last sentence of the instructions already has the right answer: "Read the
outcome field, it already says which."

Fix: cut the mechanism from both, keep the vocabulary.

```python
INSTRUCTIONS = (
    "Tools for controlling Robotiq 2F adaptive grippers. Call gripper_list_grippers "
    "to discover available names, and pass one as gripper_name to every other tool. "
    "Openings are in millimetres: 0.0 is fully closed; fully open is the model's max "
    "opening, reported by gripper_list_grippers. Every motion returns an outcome "
    "field that says how it ended; read that rather than inferring from the flags."
)
```

and in `gripper_grasp`'s description, keep the two outcome names and drop the
`reached_goal`/`stalled` explanation. One home per explanation, and it is
`classify`.

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,240 @@
...
+        ),
+    )
+    def gripper_set_opening(
+        gripper_name: str, opening_mm: float
+    ) -> GripperMotionResult:
+        return service.move_to_opening(gripper_name, opening_mm)
+
+    @mcp.tool(
+        name="gripper_grasp",
+        annotations=CLOSES,
```
ebarnett3, line 179: ### 3. `gripper_grasp` is annotated idempotent — Low

`server.py:181-183` reuses `CLOSES`, which sets `idempotent_hint: True`
(`server.py:56-61`). Repeating a grasp on something already held re-drives the
fingers into it, and a client that treats the hint as licence to retry a
timed-out call will squeeze twice. `gripper_open` and `gripper_close` are
genuinely idempotent; a grasp is not.

Fix: a `GRASPS` dict, same as `CLOSES` with `idempotent_hint: False`.

mcp/gripper_mcp/server.py
ebarnett3: I think `move_to` is more clear. With `gripper_set_opening`, it only becomes clear that the gripper will be moved after reading the description.

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,240 @@
...
+    )
+    def gripper_close(gripper_name: str) -> GripperMotionResult:
+        return service.close_fully(gripper_name)
+
+    @mcp.tool(
+        name="gripper_set_opening",
+        annotations=CLOSES,
+        description=describe(
+            """
+        Move one gripper to a specific opening.
```
ebarnett3: I would use `position` rather than `opening`

mcp/gripper_mcp/server.py
```diff
@@ -0,0 +1,240 @@
...
+        outcome="closed_without_object".
+
+        Args:
+            gripper_name: Name of the gripper (see gripper_list_grippers).
+            max_effort_n: Grip force ceiling in newtons. Omit for the model's
+                default. Lower it for fragile objects.
+            """
+        ),
+    )
+    def gripper_grasp(
```
ebarnett3: Not really sure it makes sense to have the two separate functions for `gripper_grasp` and `gripper_close` ?

Note that it is possible to grasp an object by opening the gripper, for example grasping a hollow cylinder on the inside wall.

The SDK isolates the different cases with an enum:

```cpp
enum class ObjectDetection : uint8_t
{
   Moving = detail::rm::kObjectMoving,
   DetectedWhileOpening = detail::rm::kObjectDetectedOpening,
   DetectedWhileClosing = detail::rm::kObjectDetectedClosing,
   AtRequestedPosition = detail::rm::kObjectAtRequestedPosition
};
```

mcp/README.md
ebarnett3: I don't think this can work yet because the fake gripper does not set gOBJ correctly.

As mentioned elsewhere, I think we need to create and use more realistic simulatedGripper in the SDK. That is where we can ensure that the behavior matches the real gripper as closely as possible. This will also avoid having each consumer define its own simulated gripper, with the different versions silently drifting apart in behavior.

This is not required for the current PR, but we should do this before putting too much effort in the consumer-level simulated grippers.

mcp/gripper_mcp/server.py
ebarnett3: In the SDK, this can be accomplished by either:
- setting the commanded position to the actual position
- lowering the rGTO bit of the ActionRequest byte

ebarnett3 (approved):

mcp/gripper_mcp/server.py
jlongvalRobotiq: Default is now `127.0.0.1`, `--host 0.0.0.0` is the deliberate flag, and the README says plainly that the tools carry no authentication and that the port needs firewalling when served wide.

mcp/README.md
jlongvalRobotiq: Un-documented rather than exposed: the README and `grippers.yaml.example` now point at ros2_control's `use_fake_hardware`, which is what the demo in #69 runs on. Agreed with both of you that the xacro arg and a realistic SDK simulated gripper are the real fix; both are on our follow-up list, outside this PR.

mcp/tests/test_server.py
jlongvalRobotiq: Function scope now; each test gets a fresh double.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Not reached yet rather than deliberate. #61 wires goal cancellation through the action client, and a `gripper_stop` tool on top of that is a follow-up once we know what the driver does on cancel. Eric's two routes (command the current position, or drop rGTO) are what the driver would need underneath; that lives outside mcp/.

mcp/gripper_mcp/server.py
jlongvalRobotiq: `GripperInfo.max_opening_mm`, filled from the datasheet stroke, as you sketched.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Your INSTRUCTIONS text, verbatim. With `gripper_grasp` folded away (thread below) the grasp description is gone too, so the outcome now has exactly one home.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Moot: `gripper_grasp` no longer exists (see the grasp/close thread). `open`, `close` and `move_to` stay idempotent.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Renamed: `gripper_move_to(gripper_name, position_mm)`.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Renamed: `gripper_move_to(gripper_name, position_mm)`.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Done as you suggest. `gripper_grasp` is gone; `max_effort_n` is an optional argument on `open`, `close` and `move_to`, and the outcome is neutral: `stopped_on_object` with `object_detected: true` in either direction, `reached` otherwise, mirroring `ObjectDetection` without the `Moving` state (a finished motion is never moving). Whether contact was wanted is the caller's judgement. The hollow-cylinder case falls out: an `open` that stops short is `stopped_on_object`.

PR #61, opened 2026-09-03
feat(mcp): ROS backend on rclpy

jlongvalRobotiq:
## Change

Seventh PR of the `mcp/` series (stacked on #60, now merged). The backend that drives the real driver, in three commits, bottom-up:

`robot_description.py` and `units.py` are a stopgap: the MCP needs millimetres and the driver speaks knuckle radians until #70 lands, so the backend reads the driver's own description to convert. Both files go away with #70 (robotiq/grippers#16).

1. **`robot_description.py`** — the command joint and its angles, read from the `robot_description` the backend is connected to, with no ROS imports. A Robotiq 2F is one actuated joint plus finger joints that `<mimic>` it, so the command joint is the one the others point at, whatever `prefix` the xacro was expanded with. Its lower limit is the open angle. The closed angle is the **driver's**, not the joint's: the hardware interface scales the register to `gripper_closed_position` (0.7929 on a 2F-85, against a limit of 0.8), which the description carries as a `<hardware>` param of the `<ros2_control>` block that lists the joint. The limit is the fallback where that block has no such param (fake hardware, topic-based sim). Reading the limit would have put a fully closed gripper at 0.75 mm open.
2. **`ros_messages.py`** — folds the goal's terminal status and result into a `BackendMotion`, and picks the advertised action type, with no ROS imports so both run in CI. The case that matters: with `allow_stalling: false` (the sim controller configs from #52) the controller **aborts** a goal that stalls and the result still says `stalled: true`. That abort is how the driver reports fingers stopped on an object, so it comes through as a finished motion with `stalled: true`, and the service (#59) reads the outcome from where the fingers ended up. Canceled goals are timeouts, rejected goals refusals.
3. **`ros_backend.py`** — `RosGripperBackend` on rclpy. One node per gripper, created inside the gripper's `namespace` from `grippers.yaml`, so `robotiq_gripper_controller/gripper_cmd`, `joint_states` and the latched `robot_description` resolve to that gripper. All nodes share one `MultiThreadedExecutor` spinning in a daemon thread; the MCP tools, which FastMCP runs in worker threads, wait on futures with the datasheet's motion timeout. Calls on one gripper are serialised by a per-instance lock, one deadline covers the whole call from goal response to result, and a goal that outlives it is cancelled with the cancel's answer put in `detail`, so a late result cannot land on the next call and a rejected cancel is visible. The action type is whatever the running controller advertises for `gripper_cmd`: `ParallelGripperCommand` (Jazzy's `parallel_gripper_action_controller`) or `GripperCommand` (Humble's `position_controllers`), read off the graph rather than guessed from the distro or from what imports, because recent `control_msgs` releases ship both types on every distro. Discovery is polled up to the server-wait budget, so a `health()` right after start-up does not report a controller absent that is merely undiscovered yet. `build_backend` imports it lazily; without a sourced ROS install a wiring entry fails at startup with a message naming rclpy. README gains a "With the driver" section.
4. **Review round** (Eric, mbegin) — folded into the three commits: `command_joint` uses `findall("joint")` so a `<ros2_control>` block emitted after the joints no longer shadows them; `measured_effort` is deleted and `force_n` stays `None` (no description exports an effort interface, and a joint effort would be a torque anyway); the result fallback is the last known `joint_states` position, never the commanded one; `rclpy.init()` is guarded on `rclpy.ok()`; `RosGraph.shutdown` runs from `main()`'s exit path.

Deliberately not in the protocol: activation, fault codes, speed. `gripper_cmd` offers none of them. Effort the hardware reports as NaN (mock components do) is surfaced as `null`, not NaN.

## Verification

- `uv run pytest`: 119 passed (26 new): mimicked joint found, prefixed cell, closed angle 0.7929 from the real-hardware expansion and 0.8 from the fake-hardware one, another joint's `ros2_control` block ignored, `<ros2_control>` emitted after the joints, arm-only / two-gripper / no-limit descriptions refused by name; succeeded, aborted-with-stall, succeeded-with-stall, canceled, refusal, timeout with the cancel accepted / rejected / unanswered, type picked from the advertised list, other namespace ignored, unknown type ignored; the missing-rclpy startup message.
- **Live, driver on ros2_control's fake hardware**, `robotiq_description` from this branch, in the repo's own `docker/Dockerfile` images:

  | Image | Model | Bound type | health | move 0.4 / 0.0 |
  |---|---|---|---|---|
  | `robotiq_ros2:jazzy` | 2F-85, 2F-140 | `ParallelGripperCommand` | ready, fresh | succeeded, `reached_goal` |
  | `robotiq_ros2:humble` | 2F-85, 2F-140 | `GripperCommand` | ready, fresh | succeeded, `reached_goal` |

  `read_state()` follows each move; `force_n` is `None` on fake hardware. Lyrical/Rolling share Jazzy's code path.
- Not yet exercised: abort-on-stall against Isaac Sim (needs the twin running). The mapping is unit-tested.

## Stack
#60 is merged; this PR shows its own 3 commits.

## Next
ros#70 (driver on the SDK's SI conversions) makes the driver speak metres; the closed-angle reading and `units.py` then go away together.

### DIFF
--- a/mcp/gripper_mcp/robot_description.py
+++ b/mcp/gripper_mcp/robot_description.py
@@ -1,0 +1,58 @@
+"""Find the gripper's command joint in a robot description, with no ROS imports.
+
+A Robotiq 2F is one actuated joint plus finger joints that `<mimic>` it, so the
+command joint is the one the others point at. That holds whatever `prefix` the
+xacro was expanded with, which is why the name is read here rather than kept in
+a datasheet. The joint's lower limit is the open angle. The closed angle is the
+driver's, not the joint's: the hardware interface scales the register to
+`gripper_closed_position` (0.7929 on a 2F-85, against a limit of 0.8), which
+the description carries as a `<hardware>` param of the `<ros2_control>` block
+that lists the joint. Where that block has no such param (the fake hardware,
+the topic-based sim) the mock mirrors commands and the limit is as good as
+anything, so it is the fallback.
+"""
+
+from xml.etree import ElementTree
+
+from gripper_mcp.units import JointGeometry
+
+
+def command_joint(urdf_xml: str) -> JointGeometry:
+    robot = ElementTree.fromstring(urdf_xml)
+    joints = {joint.attrib["name"]: joint for joint in robot.findall("joint")}
+    name = mimicked_joint(robot, set(joints))
+    rad_open, limit_upper = limits(joints[name])
+    return JointGeometry(name, rad_open, closed_angle(robot, name, limit_upper))
+
+
+def mimicked_joint(robot: ElementTree.Element, joint_names: set[str]) -> str:
+    mimicked = sorted(
+        {mimic.attrib["joint"] for mimic in robot.iter("mimic")} & joint_names
+    )
+    if len(mimicked) != 1:
+        raise ValueError(
+            "Expected exactly one joint mimicked by the finger joints, found "
+            f"{len(mimicked)}: {', '.join(mimicked) or '(none)'}. Is this the "
+            "description of a single Robotiq 2F gripper?"
+        )
+    return mimicked[0]
+
+
+def closed_angle(robot: ElementTree.Element, joint: str, limit_upper: float) -> float:
+    for block in robot.iter("ros2_control"):
+        if block.find(f"joint[@name='{joint}']") is None:
+            continue
+        param = block.find("hardware/param[@name='gripper_closed_position']")
+        if param is not None and param.text:
+            return float(param.text)
+    return limit_upper
+
+
+def limits(joint: ElementTree.Element) -> tuple[float, float]:
+    limit = joint.find("limit")
+    try:
+        return float(limit.attrib["lower"]), float(limit.attrib["upper"])
+    except (AttributeError, KeyError) as error:
+        raise ValueError(
+            f"Joint '{joint.attrib['name']}' has no <limit lower= upper=> in the URDF"
+        ) from error
--- a/mcp/gripper_mcp/ros_backend.py
+++ b/mcp/gripper_mcp/ros_backend.py
@@ -1,0 +1,311 @@
+"""Gripper backend on the ROS 2 driver, through rclpy.
+
+One node per gripper, created inside the gripper's namespace so that
+`robotiq_gripper_controller/gripper_cmd`, `joint_states` and `robot_description`
+resolve to that gripper's controller. All nodes share one executor spinning in a
+background thread; the MCP tools run in their own threads and wait on futures.
+
+The command joint's name and range are read from `robot_description`, the URDF
+robot_state_publisher latches for the cell, so a prefixed cell resolves by itself
+and nothing is copied from a datasheet. It is resolved on first use and kept.
+
+The action type is whatever the running controller advertises for `gripper_cmd`:
+`ParallelGripperCommand` from Jazzy's parallel_gripper_action_controller,
+`GripperCommand` from Humble's position_controllers one. It is read off the graph
+rather than guessed from the distro or from what imports, because recent
+control_msgs releases ship both types on every distro.
+
+Calls on one gripper are serialised: FastMCP runs tools on worker threads and an
+agent may issue two moves at once, and two goals in flight on one controller
+resolve in whatever order it likes. One deadline covers the whole call, from goal
+response to result, so the tool-level ceiling is the datasheet's timeout. A goal
+that outlives it is cancelled and the cancel's answer is reported, because a
+rejected cancel means the gripper is still moving after the tool said it failed.
+
+`force_n` is never filled. `joint_states.effort` would be a joint torque in N m,
+and no shipped description exports an effort interface anyway; the gripper
+reports motor current (gCU), and turning that into a fingertip force needs a
+calibration nobody has done. The commanded force is the honest number.
+"""
+
+import threading
+import time
+from collections.abc import Callable
+
+import rclpy
+from control_msgs.action import GripperCommand, ParallelGripperCommand
+from rclpy.action import ActionClient
+from rclpy.action.graph import get_action_names_and_types
+from rclpy.executors import MultiThreadedExecutor
+from rclpy.node import Node
+from rclpy.qos import DurabilityPolicy, QoSProfile
+from sensor_msgs.msg import JointState
+from std_msgs.msg import String
+
+from gripper_mcp.backend import BackendHealth, BackendMotion, BackendState
+from gripper_mcp.robot_description import command_joint
+from gripper_mcp.units import JointGeometry
+from gripper_mcp.ros_messages import (
+    advertised_type,
+    cancel_note,
+    motion_from_result,
+    position_of,
+    refused,
+    timed_out,
+)
+
+ACTION_NAME = "robotiq_gripper_controller/gripper_cmd"
+ACTION_TYPES = {
+    "control_msgs/action/ParallelGripperCommand": ParallelGripperCommand,
+    "control_msgs/action/GripperCommand": GripperCommand,
+}
+JOINT_STATES_TOPIC = "joint_states"
+DESCRIPTION_TOPIC = "robot_description"
+LATCHED = QoSProfile(depth=1, durability=DurabilityPolicy.TRANSIENT_LOCAL)
+SERVER_WAIT_S = 2.0
+DESCRIPTION_WAIT_S = 2.0
+STATE_WAIT_S = 2.0
+STALE_STATE_S = 2.0
+POLL_S = 0.05
+
+
+class RosGraph:
+    _lock = threading.Lock()
+    _executor: MultiThreadedExecutor | None = None
+
+    @classmethod
+    def executor(cls) -> MultiThreadedExecutor:
+        with cls._lock:
+            if cls._executor is None:
+                if not rclpy.ok():
+                    rclpy.init()
+                cls._executor = MultiThreadedExecutor()
+                threading.Thread(
+                    target=cls._executor.spin, name="rclpy-spin", daemon=True
+                ).start()
+            return cls._executor
+
+    @classmethod
+    def shutdown(cls) -> None:
+        with cls._lock:
+            if cls._executor is not None:
+                cls._executor.shutdown()
+                rclpy.shutdown()
+                cls._executor = None
+
+
+def wait_for(future, timeout_s: float):
+    done = threading.Event()
+    future.add_done_callback(lambda _: done.set())
+    if not done.wait(timeout_s):
+        return None
+    return future.result()
+
+
+def poll_until(probe: Callable[[], object], timeout_s: float):
+    deadline = time.monotonic() + timeout_s
+    while True:
+        found = probe()
+        if found is not None or time.monotonic() >= deadline:
+            return found
+        time.sleep(POLL_S)
+
+
+def remaining(deadline: float) -> float:
+    return max(0.0, deadline - time.monotonic())
+
+
+def cancel_when_accepted(goal_future) -> None:
+    def cancel(done) -> None:
+        handle = done.result()
+        if handle.accepted:
+            handle.cancel_goal_async()
+
+    goal_future.add_done_callback(cancel)
+
+
+class JointStateFeed:
+    def __init__(self, node: Node) -> None:
+        self.latest: JointState | None = None
+        self.latest_at = 0.0
+        self._joint_name: str | None = None
+        node.create_subscription(
+            JointState, JOINT_STATES_TOPIC, self._on_joint_states, 10
+        )
+
+    def follow(self, joint_name: str) -> None:
+        self._joint_name = joint_name
+
+    def _on_joint_states(self, message: JointState) -> None:
+        if self._joint_name is not None and self._joint_name in message.name:
+            self.latest = message
+            self.latest_at = time.monotonic()
+
+
+class RosGripperBackend:
+    name = "ros"
+
+    def __init__(self, gripper_name: str, namespace: str) -> None:
+        executor = RosGraph.executor()
+        self._node = Node(f"gripper_mcp_{gripper_name}", namespace=namespace or "/")
+        self._description: str | None = None
+        self._node.create_subscription(
+            String, DESCRIPTION_TOPIC, self._on_description, LATCHED
+        )
+        self._states = JointStateFeed(self._node)
+        self._joint: JointGeometry | None = None
+        self._client: ActionClient | None = None
+        self._action_type = None
+        self._client_lock = threading.Lock()
+        self._motion_lock = threading.Lock()
+        executor.add_node(self._node)
+
+    def joint_geometry(self) -> JointGeometry:
+        if self._joint is None:
+            if not self._await_description():
+                raise RuntimeError(
+                    f"No {DESCRIPTION_TOPIC} under '{self._node.get_namespace()}' "
+                    f"within {DESCRIPTION_WAIT_S:.0f} s; is robot_state_publisher "
+                    "running there?"
+                )
+            self._joint = command_joint(self._description)
+            self._states.follow(self._joint.name)
+        return self._joint
+
+    def read_state(self) -> BackendState:
+        joint = self.joint_geometry().name
+        if not self._await_state():
+            raise RuntimeError(
+                f"No {JOINT_STATES_TOPIC} naming '{joint}' under "
+                f"'{self._node.get_namespace()}' within {STATE_WAIT_S:.0f} s; "
+                "is the controller running?"
+            )
+        return BackendState(position_rad=position_of(self._states.latest, joint, 0.0))
+
+    def move_to(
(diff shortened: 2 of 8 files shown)


mcp/gripper_mcp/ros_backend.py
```diff
@@ -0,0 +1,284 @@
...
+        if wrapped is None:
+            handle.cancel_goal_async()
+            return timed_out(self._position_or(here), timeout_s, "result")
+
+        return motion_from_result(
+            wrapped.status,
+            self._result_position(wrapped.result, position_rad),
+            wrapped.result.stalled,
+            wrapped.result.reached_goal,
+        )
```
ebarnett3, line 227: ### 1. `move_to` re-derives what a finished move is, for the third time — Design

`move_to` (`ros_backend.py:163-195`) decides when a motion is over and what it
meant, from the action's terminal status and its `stalled` / `reached_goal`
fields. The gripper already answers both questions, and the SDK already has the
sequence that asks it, in `examples/move_gripper.cpp`: wait for the position
echo, optionally observe `Moving`, wait for `motionSettled`, then read the gOBJ
verdict. That is the device's own completion logic, and it sits in an example
rather than in the public API, so every consumer writes it again. Three copies
now: that example, `parallel_gripper_action_controller`'s velocity-based stall
detection, and this backend plus #59's classifier.

The moveTo function should be moved to the public API in the SDK, along with the interpretation of the move result.

Nothing to change in this PR on its own: the backend can only read what the
graph offers, and today that is the action result. Filed here because this is
where the third copy appears, and because the fix is an SDK issue plus a ros
issue rather than a patch to `ros_backend.py`. Once both land, `move_to` sends
a goal, waits for the future and reports the verdict, and #59's `classify`
mostly disappears.

mcp/gripper_mcp/ros_backend.py
```diff
@@ -0,0 +1,284 @@
...
+        if client is None:
+            return refused(
+                here,
+                f"No {ACTION_NAME} action server under "
+                f"'{self._node.get_namespace()}'; is the controller active?",
+            )
+
+        goal = self._goal(position_rad, max_effort_n)
+        sent = client.send_goal_async(goal)
+        handle = wait_for(sent, timeout_s)
```
ebarnett3: ### 2. A motion can block for more than twice its timeout — Medium

`move_to` waits `timeout_s` for the goal response (`ros_backend.py:178`) and
`timeout_s` again for the result (`ros_backend.py:185`), so the datasheet's
10 s budget can hold a tool for 20 s, plus up to two `SERVER_WAIT_S` in
`_ready_client` (`ros_backend.py:246-251`). One deadline for the call fixes it:

```python
deadline = time.monotonic() + timeout_s
handle = wait_for(sent, max(0.0, deadline - time.monotonic()))
...
wrapped = wait_for(handle.get_result_async(), max(0.0, deadline - time.monotonic()))
```

mcp/gripper_mcp/ros_backend.py
```diff
@@ -0,0 +1,284 @@
...
+        if handle.accepted:
+            handle.cancel_goal_async()
+
+    goal_future.add_done_callback(cancel)
+
+
+def measured_effort(efforts: Sequence[float], index: int) -> float | None:
+    if len(efforts) <= index or math.isnan(efforts[index]):
+        return None
+    return efforts[index]
```
ebarnett3: ### 3. `measured_effort` reads an interface no description exports — Low

`measured_effort` (`ros_backend.py:111-114`) reads `joint_states.effort` into
`BackendState.force_n`, which the service publishes as `GripperState.force_n`,
"measured grip force in newtons". No configuration exports that interface:
`2f_85.ros2_control.xacro:56-63` declares `position` and `velocity` in every
branch and adds `motor_current` and `object_status` only for the real driver.
So the list is always empty and the function always returns `None`. It is dead
code whose only reachable effect is to publish the wrong unit if someone later
adds the interface, since a revolute joint's effort is a torque in N·m.

Renaming the field would not rescue it, because there is no accurate torque to
publish either. The gripper reports gCU, motor current at 10 mA per count.
Deriving a fingertip force from that needs a motor constant, a gear ratio, the
four-bar geometry and friction, and the manual gives grip force against the
force-request setting rather than a current-to-force curve. Any number produced
that way is an estimate nobody has calibrated.

Fix: delete the read, leave `force_n` `None`, and say in `read_state` that the
driver publishes no force. If a force reading is wanted later, the two honest
sources are the commanded force, which the service already knows, and
`motor_current` once a broadcaster publishes it, reported in its own unit
rather than converted.

mcp/gripper_mcp/ros_backend.py
```diff
@@ -0,0 +1,284 @@
...
+}
+JOINT_STATES_TOPIC = "joint_states"
+DESCRIPTION_TOPIC = "robot_description"
+LATCHED = QoSProfile(depth=1, durability=DurabilityPolicy.TRANSIENT_LOCAL)
+SERVER_WAIT_S = 2.0
+DESCRIPTION_WAIT_S = 2.0
+STATE_WAIT_S = 2.0
+STALE_STATE_S = 2.0
+POLL_S = 0.05
+CLIENT_LOCK = threading.Lock()
```
ebarnett3: ### 4. `CLIENT_LOCK` is module-global, so one absent controller stalls the others — Low

`ros_backend.py:58` and `:244`. The lock is shared by every gripper's backend and
is held across `poll_until(self._advertised_type, SERVER_WAIT_S)`, so a cell
whose left gripper has no controller makes every call on the right gripper wait
out the left one's 2 s poll. Fix: make it an instance lock in `__init__`, which
is all the invariant needs.

mcp/gripper_mcp/ros_backend.py
```diff
@@ -0,0 +1,284 @@
...
+                ).start()
+            return cls._executor
+
+    @classmethod
+    def shutdown(cls) -> None:
+        with cls._lock:
+            if cls._executor is not None:
+                cls._executor.shutdown()
+                rclpy.shutdown()
+                cls._executor = None
```
ebarnett3, line 94: ### 5. `RosGraph.shutdown()` is never called — Nit

`ros_backend.py:76-82` has no caller anywhere in the tree. Either wire it into
`main()`'s exit path or drop it; as written it reads as a lifecycle that exists
and is not used.

mcp/gripper_mcp/ros_backend.py
```diff
@@ -0,0 +1,284 @@
...
+
+class RosGraph:
+    _lock = threading.Lock()
+    _executor: MultiThreadedExecutor | None = None
+
+    @classmethod
+    def executor(cls) -> MultiThreadedExecutor:
+        with cls._lock:
+            if cls._executor is None:
+                rclpy.init()
```
ebarnett3: ### 6. `RosGraph.executor()` calls `rclpy.init()` unconditionally — Nit

`ros_backend.py:69`. Fine for the server process, but it raises if the module is
ever imported into a host that already initialised rclpy. `rclpy.ok()` guards it
in one line.

mcp/gripper_mcp/robot_description.py
```diff
@@ -0,0 +1,58 @@
...
+"""
+
+from xml.etree import ElementTree
+
+from gripper_mcp.units import JointGeometry
+
+
+def command_joint(urdf_xml: str) -> JointGeometry:
+    robot = ElementTree.fromstring(urdf_xml)
+    joints = {joint.attrib["name"]: joint for joint in robot.iter("joint")}
```
mbegin-robotiq: `.iter()` recurses, so this also matches the `<joint name="...">` elements inside `<ros2_control>` — which carry no `<limit>`. Whichever appears last in document order wins the dict.

It works today only by luck of emission order: `robotiq_2f_85_macro.urdf.xacro` emits `<ros2_control>` at line 41 and the real joints at line 256+, so the real one overwrites the stub. Put the `<ros2_control>` block after the joints — perfectly legal URDF — and `command_joint` raises:

```
ValueError: Joint 'left_knuckle_joint' has no <limit lower= upper=> in the URDF
```

on a joint that plainly has one. Silent order-dependence, misleading error.

`robot.findall("joint")` fixes it: URDF joints are always direct children of `<robot>`, ros2_control's are nested. Verified against a live graph — with that one word changed the description parses correctly, `JointGeometry(name='left_knuckle_joint', rad_open=0.0, rad_closed=0.7929)`, closed angle read from the hardware param as intended.

Worth a test with `<ros2_control>` emitted last, since every current fixture happens to match the shipped xacro's ordering.

mcp/gripper_mcp/ros_backend.py
mbegin-robotiq: The description says *"Calls are serialised per gripper and a goal that outlives its deadline is cancelled, so a late result cannot land on the next call."* Neither half holds — this is the first.

`CLIENT_LOCK` is module-global and only guards action-client construction; nothing serialises `move_to`. Three concurrent calls on one gripper, against a fake controller taking 1 s per goal:

```
wall clock, 3 concurrent move_to on ONE gripper: 1.02 s
per-call durations: [(0.2, 1.01), (0.4, 1.01), (0.6, 1.01)]
goals accepted: 3 | max concurrent in server: 3
```

All three in flight at once, three conflicting positions, undefined resolution order. FastMCP runs tools in worker threads and agents routinely issue parallel tool calls, so this is reachable in normal use. A per-instance lock around `move_to` would close it.

Separately, this lock being module-global means the first call per gripper holds it through up to `SERVER_WAIT_S` of discovery polling, so N grippers serialise into N x 2 s on first touch. Per-instance fixes both.

mcp/gripper_mcp/ros_backend.py
```diff
@@ -0,0 +1,284 @@
...
+        handle = wait_for(sent, timeout_s)
+        if handle is None:
+            cancel_when_accepted(sent)
+            return timed_out(here, timeout_s, "goal response")
+        if not handle.accepted:
+            return refused(here, "The controller rejected the goal.")
+
+        wrapped = wait_for(handle.get_result_async(), timeout_s)
+        if wrapped is None:
+            handle.cancel_goal_async()
```
mbegin-robotiq: The second half of the serialisation claim. The cancel is fired and its result never checked, so a rejected cancel is indistinguishable from an honoured one.

Against a fake controller slower than the deadline:

```
call 1 (timeout_s=1, controller takes 3s): timed_out=True | No result from the controller within 1.0 s.
  gripper position AFTER the abandoned goal finished: 0.75   <- it still moved
```

`move_to` returned `incomplete` and the gripper carried on to the commanded position. rclpy action servers reject cancel requests by default unless a `cancel_callback` accepts them.

Caveat: this is my own fake server, not `parallel_gripper_action_controller`, which may well honour cancel — so treat this as unverified against the real controller rather than broken. The point that stands either way is that the backend never checks, so it cannot know, and reporting a failure while the hardware keeps moving is the worst direction for this to fail in. Awaiting the cancel response (or at least logging a rejected one) would make it observable.

mcp/gripper_mcp/ros_backend.py
```diff
@@ -0,0 +1,284 @@
...
+        if self._action_type is ParallelGripperCommand:
+            goal.command.name = [self._joint.name]
+            goal.command.position = [position_rad]
+            goal.command.effort = [max_effort_n]
+        else:
+            goal.command.position = position_rad
+            goal.command.max_effort = max_effort_n
+        return goal
+
+    def _result_position(self, result, fallback: float) -> float:
```
mbegin-robotiq, line 308: The fallback here is the *commanded* position, so a result that cannot report where the fingers actually are reads as though they arrived.

On a `ParallelGripperCommand` whose `result.state` omits the joint, a grasp targeting `min_opening_mm` reports achieved == commanded == 0 mm, `stopped_on_something` goes false, and a successful grasp classifies as `closed_without_object` — the exact misreading the outcome vocabulary exists to prevent.

`here` (last known `joint_states` position, computed at the top of `move_to`) is the honest fallback and is already to hand. Not reproduced — `parallel_gripper_action_controller` does normally populate `state` — but the failure is silent and lands on the field agents are told to trust.

mcp/gripper_mcp/ros_backend.py
```diff
@@ -0,0 +1,284 @@
...
+        goal = self._goal(position_rad, max_effort_n)
+        sent = client.send_goal_async(goal)
+        handle = wait_for(sent, timeout_s)
+        if handle is None:
+            cancel_when_accepted(sent)
+            return timed_out(here, timeout_s, "goal response")
+        if not handle.accepted:
+            return refused(here, "The controller rejected the goal.")
+
+        wrapped = wait_for(handle.get_result_async(), timeout_s)
```
mbegin-robotiq: `timeout_s` is spent twice: once waiting for the goal response above, then again here for the result. Add `_ready_client`'s up to 2 s of discovery polling plus 2 s of `wait_for_server` and a single `gripper_grasp` can occupy a FastMCP worker for roughly 24 s against the datasheet's `motion_timeout_s: 10.0`.

Either share one deadline across the phases, or document that the tool-level ceiling is a multiple of the datasheet value — an agent or client with its own timeout will otherwise give up first.

mcp/gripper_mcp/ros_backend.py
mbegin-robotiq: `rclpy.init()` unconditionally raises if the host process has already initialised rclpy — which is the normal case if the MCP is ever embedded in a node, and it also makes this awkward to drive from a test that owns its own context.

`rclpy.ok()` guard, or accepting an injected executor, would make the backend usable in both. Related: `RosGraph.shutdown()` exists but nothing calls it, so nodes and the executor thread outlive the server's own lifecycle.

mbegin-robotiq (approved):

ebarnett3 (approved):

mcp/gripper_mcp/ros_backend.py
jlongvalRobotiq: Agreed, and it is the change that makes most of `classify` disappear. Nothing changes here for it. Filing it against the SDK (moveTo plus the verdict in the public API) and against this repo for the driver surfacing gOBJ; `move_to` shrinks to send, wait, report once both land.

mcp/gripper_mcp/ros_backend.py
jlongvalRobotiq: One deadline now, shared by the goal-response wait and the result wait, so the tool-level ceiling is the datasheet's `motion_timeout_s`. The discovery wait in `_ready_client` stays outside it; it only happens on the first call per gripper.

mcp/gripper_mcp/ros_backend.py
jlongvalRobotiq: Deleted. `force_n` stays `None` and the module docstring says why: no description exports an effort interface, and a revolute joint's effort would be a torque anyway. On the commanded side, grippers#36 settles that rFR is a unitless effort in [0, 1] rather than newtons, so `max_effort_n` follows it in the ros#70 units track; measured force stays absent until `motor_current` is published in its own unit.

mcp/gripper_mcp/ros_backend.py
jlongvalRobotiq: Wired: `main()` runs the registered shutdown hooks in a `finally` around `mcp.run`, so the nodes and the spin thread end with the server.

mcp/gripper_mcp/robot_description.py
jlongvalRobotiq: `findall("joint")` now, plus a test whose fixture emits `<ros2_control>` after the joints. Good catch; it worked by emission order only.

mcp/gripper_mcp/ros_backend.py
jlongvalRobotiq: Per-instance lock around the body of `move_to`, and the client lock is per-instance too, so N grippers no longer serialise their discovery. The serialisation itself has no unit test because rclpy is not importable in the test venv; it is on the bench list.

mcp/gripper_mcp/ros_backend.py
jlongvalRobotiq: The result-timeout path awaits the cancel response, bounded, and puts a rejected or unanswered cancel in `detail`, so an `incomplete` while the hardware keeps moving is at least visible. Whether `parallel_gripper_action_controller` honours cancel is on the bench list.

mcp/gripper_mcp/ros_backend.py
jlongvalRobotiq: Fallback is the last known `joint_states` position now, never the commanded one.

mcp/gripper_mcp/ros_backend.py
jlongvalRobotiq: One deadline across goal response and result (Eric's thread above); the ceiling is the datasheet value plus first-call discovery.

mcp/gripper_mcp/ros_backend.py
jlongvalRobotiq: Guarded on `rclpy.ok()`, and `RosGraph.shutdown` now runs from `main()`'s exit path. Injected executor not done; the guard covers the embedded case.

PR #62, opened 2026-09-03
feat(mcp): tactile protocol, baseline maths and mock tactile source

jlongvalRobotiq:
## Change

Eighth PR of the `mcp/` series (stacked on #61, now merged). The tactile domain, with no ROS and no tool surface yet:

1. **`tactile_backend.py`** — a second, optional protocol next to `GripperBackend`: `TactileBackend.read_tactile() -> TactileReading`, where a reading is one `TactilePad` per finger (raw `uint16` counts, exactly as `robotiq_tsf/msg/StaticData` publishes them) and a `TactileLayout` (7x4, two pads for the TSF-85). It is separate from the gripper protocol because the pads are their own device with their own driver, a bare 2F has none, and a simulator may model fingers without modelling touch.
2. **`tactile.py`** — the one place raw counts become meaningful: `average_readings` (mean of many samples, what the SDK's `findBaseline()` does over 1000), `pad_sums`/`peak_rise`/`contact_signal` over baseline-subtracted rises. `contact_signal` is normalised across both pads: 1.0 means every taxel on both pads at the working ceiling, so a firm one-sided contact reads about half; thresholds are in those units. Pads are matched to their baseline by name, never by position, so a source that reorders its pads cannot have one pad's rest subtracted from the other's counts. Rises clamp at zero per taxel, so a pad drifting below its rest cannot cancel real contact on the other one. Shape mismatches raise `TactileShapeMismatch` instead of silently zipping short; an empty sample list is a plain `ValueError`.
3. **`tests/fakes/tactile.py`** — the test double for the tactile tests, next to the fake gripper (#57). Pressure follows how far the fingers have closed past a virtual object, read live through a callable so it tracks the fake gripper it sits beside, and every taxel of a pad rises by the same amount, with a light touch registered the moment the fingers reach the object (the mock gripper stalls exactly there, and real pads are soft); its object width is independent of the gripper's on purpose, so "stalled but the pads stayed quiet" is reachable in a test. It does not ship: without TSF pads there is nothing for the tactile tools to read, and a production stand-in inventing a signal would be worse than reporting none.

Nothing in `service.py`/`server.py` changes yet: the tactile datasheet block, `grippers.yaml` wiring, the `gripper_read_tactile`/`gripper_tare_tactile`/`gripper_verify_grasp` tools and the ROS source on `TactileSensor/StaticData` are the next PRs. The pad's `Dynamic` channel is not carried: nothing in the series reads it (Eric's review), so it comes back with its first consumer.

## Verification

- `uv run pytest`: 147 passed (28 new): baseline mean, baseline keyed by pad name, samples averaged by name with pads swapped, empty list is a `ValueError`, shape errors, signal 0 / 0.5 / 1.0, non-positive full scale is a `ValueError`, clamp-at-zero and no-masking, swapped pads still subtract the right rest, a pad the baseline never saw is refused, peak rise, wrong-shape baseline; fake at rest / light touch on reaching the object / pressed / saturated, both pads agree, layout, live opening tracking, protocol conformance.

## Stack
#61 is merged; this PR shows its own 3 commits.

### DIFF
--- a/mcp/gripper_mcp/tactile.py
+++ b/mcp/gripper_mcp/tactile.py
@@ -1,0 +1,122 @@
+"""Baseline subtraction and contact-signal arithmetic over raw taxel counts.
+
+Raw counts carry a large, drifting rest offset, so nothing is meaningful until a
+baseline is subtracted. The baseline is the mean of many samples, matching the
+tactile SDK's `findBaseline()`, which averages 1000.
+
+Pads are matched to their baseline by name, never by position, so a source that
+reports its pads in a different order cannot have one pad's rest subtracted from
+the other's counts.
+
+Per-taxel rises clamp at zero. Without that clamp a pad reading below its
+baseline, ordinary drift, would subtract from the contact detected on the other
+pad, and a firm one-sided grasp could report no contact at all.
+
+`contact_signal` is normalised across both pads: 1.0 means every taxel on both
+pads sits at the working ceiling, so a firm one-sided contact reads about half
+of what the same force spread over both pads reads. Thresholds against it are
+in those units.
+"""
+
+from dataclasses import dataclass
+
+from gripper_mcp.tactile_backend import TactilePad, TactileReading
+
+
+class TactileShapeMismatch(Exception):
+    pass
+
+
+@dataclass(frozen=True)
+class TactileBaseline:
+    pad_names: tuple[str, ...]
+    pads: tuple[tuple[float, ...], ...]
+
+    def rests_for(self, pad_name: str) -> tuple[float, ...]:
+        return self.pads[self.pad_names.index(pad_name)]
+
+
+def average_readings(readings: list[TactileReading]) -> TactileBaseline:
+    if not readings:
+        raise ValueError("Cannot average an empty list of tactile readings.")
+
+    first = readings[0]
+    for reading in readings:
+        _require_same_shape(reading, first)
+
+    pad_names = tuple(pad.name for pad in first.pads)
+
+    return TactileBaseline(
+        pad_names=pad_names,
+        pads=tuple(_average_pad(readings, name) for name in pad_names),
+    )
+
+
+def pad_sums(reading: TactileReading, baseline: TactileBaseline) -> tuple[float, ...]:
+    return tuple(sum(rises) for rises in _rises(reading, baseline))
+
+
+def peak_rise(reading: TactileReading, baseline: TactileBaseline) -> float:
+    return max(
+        (rise for rises in _rises(reading, baseline) for rise in rises), default=0.0
+    )
+
+
+def contact_signal(
+    reading: TactileReading, baseline: TactileBaseline, full_scale_counts: float
+) -> float:
+    if full_scale_counts <= 0.0:
+        raise ValueError(
+            f"full_scale_counts must be positive, got {full_scale_counts}."
+        )
+
+    return sum(pad_sums(reading, baseline)) / full_scale_counts
+
+
+def _average_pad(readings: list[TactileReading], pad_name: str) -> tuple[float, ...]:
+    samples = zip(*(_pad(reading, pad_name).taxels for reading in readings))
+
+    return tuple(sum(taxels) / len(readings) for taxels in samples)
+
+
+def _rises(
+    reading: TactileReading, baseline: TactileBaseline
+) -> tuple[tuple[float, ...], ...]:
+    _require_matching_baseline(reading, baseline)
+
+    return tuple(
+        tuple(
+            max(0.0, taxel - rest)
+            for taxel, rest in zip(pad.taxels, baseline.rests_for(pad.name))
+        )
+        for pad in reading.pads
+    )
+
+
+def _pad(reading: TactileReading, pad_name: str) -> TactilePad:
+    return next(pad for pad in reading.pads if pad.name == pad_name)
+
+
+def _require_same_shape(reading: TactileReading, other: TactileReading) -> None:
+    same = reading.layout == other.layout and _taxel_counts(reading) == _taxel_counts(
+        other
+    )
+    if not same:
+        raise TactileShapeMismatch(
+            "Tactile readings differ in shape; they cannot be averaged together."
+        )
+
+
+def _require_matching_baseline(
+    reading: TactileReading, baseline: TactileBaseline
+) -> None:
+    covered = dict(zip(baseline.pad_names, (len(rests) for rests in baseline.pads)))
+    if _taxel_counts(reading) != covered:
+        raise TactileShapeMismatch(
+            f"Baseline covers pads {sorted(covered)} but the reading has "
+            f"{sorted(_taxel_counts(reading))}; re-tare before reading."
+        )
+
+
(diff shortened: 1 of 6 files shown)


mcp/gripper_mcp/tactile.py
```diff
@@ -0,0 +1,100 @@
+"""Baseline subtraction and contact-signal arithmetic over raw taxel counts.
+
+Raw counts carry a large, drifting rest offset, so nothing is meaningful until a
+baseline is subtracted. The baseline is the mean of many samples, matching the
+tactile SDK's `findBaseline()`, which averages 1000.
```
mbegin-robotiq, line 5: We have that logic embeded in many different places now. we should make an effort to centralize

jlongvalRobotiq: Agreed. Counting them: the tactile SDK's `findBaseline()`, `tactile_total_sum.cpp`, `tactile_viz_node` (with its `rezero`) and now this file all take a mean over N frames at rest and clamp the rise at zero.

The place to centralize is the driver, not a shared library: `poll_data_sdk` already holds the SDK's `findBaseline()`, so it can expose a `tare` service and publish a baseline-subtracted `TactileSensor/Rises` topic. Then the two nodes and the MCP consume rises and delete their own baseline code; this module shrinks to `contact_signal`. Opened as #72 so it gets its own review; keeping the maths here until the driver publishes rises, otherwise the MCP has nothing to read.

mcp/gripper_mcp/tactile_backend.py
```diff
@@ -0,0 +1,46 @@
...
+    @property
+    def taxels_per_pad(self) -> int:
+        return self.rows * self.cols
+
+
+@dataclass(frozen=True)
+class TactilePad:
+    name: str
+    taxels: tuple[int, ...]
+    dynamic: int = 0
```
ebarnett3: ### 1. `TactilePad.dynamic` is carried through the protocol and never read — Low

`tactile_backend.py:33` adds the field, #66 fills it from
`TactileSensor/Dynamic`, and nothing consumes it: at the tip of the train the
only readers are two assertions in `tests/test_ros_tactile_messages.py`. So the
tactile source subscribes to a second topic, and the reading carries a value,
purely so a test can check it round-trips.

Either drop it until a tool reports it, which keeps the protocol honest and lets
#66 drop the `Dynamic` subscription, or state in the docstring what it is for
(first-touch detection is the obvious consumer, since the dynamic channel is the
pressure-change signal) so the next reader knows why it is unused. As it stands
`TactileReading` has a field that means nothing to any caller.

mcp/gripper_mcp/tactile.py
```diff
@@ -0,0 +1,117 @@
...
+    pad_names: tuple[str, ...]
+    pads: tuple[tuple[float, ...], ...]
+
+    def rests_for(self, pad_name: str) -> tuple[float, ...]:
+        return self.pads[self.pad_names.index(pad_name)]
+
+
+def average_readings(readings: list[TactileReading]) -> TactileBaseline:
+    if not readings:
+        raise TactileShapeMismatch("Cannot average an empty list of tactile readings.")
```
ebarnett3: ###  An empty reading list is reported as a shape mismatch — Nit

`tactile.py:34-36`. `average_readings([])` raises `TactileShapeMismatch`
("Cannot average an empty list of tactile readings"), which is not a shape
mismatch and reads oddly in a caller that catches that type to mean "re-tare".
`ValueError` fits, and the one call site (#65's `_capture_baseline`) can only
reach it with `baseline_samples <= 0`, which `TactileSpec` could reject instead
with `Field(gt=0)`.

mcp/gripper_mcp/tactile.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def pad_sums(reading: TactileReading, baseline: TactileBaseline) -> tuple[float, ...]:
+    return tuple(sum(rises) for rises in _rises(reading, baseline))
+
+
+def peak_taxel(reading: TactileReading, baseline: TactileBaseline) -> float:
+    return max(
+        (rise for rises in _rises(reading, baseline) for rise in rises), default=0.0
+    )
```
ebarnett3, line 62: ### `peak_taxel` returns a rise, not a taxel — Nit

`tactile.py:54-57` returns the largest per-taxel rise. #65 publishes it as
`peak_taxel_counts`, which is right; the function name reads like an index or a
pad location. `peak_rise` says what it computes.

mcp/gripper_mcp/tactile.py
```diff
@@ -0,0 +1,117 @@
...
+
+def contact_signal(
+    reading: TactileReading, baseline: TactileBaseline, full_scale_counts: float
+) -> float:
+    if full_scale_counts <= 0.0:
+        raise ValueError(
+            f"full_scale_counts must be positive, got {full_scale_counts}."
+        )
+
+    return sum(pad_sums(reading, baseline)) / full_scale_counts
```
ebarnett3, line 73: ### `contact_signal` sums both pads against one full-scale figure — Low (surfaces in #65)

`tactile.py:60-68` divides the sum over both pads by a single
`full_scale_counts`. That is a defensible definition, but it means a firm
one-sided contact reads about half of what the same force reads spread over
both pads, and #65's per-pad values divide by the same two-pad figure
(`tactile_service.py:140`), so a saturated single pad reports 0.5. Worth one
sentence in this docstring saying the signal is normalised across both pads, so
the threshold in the datasheet is read in the right units. Filed against #65 as
well, where the numbers are consumed.

ebarnett3 (approved):

mcp/gripper_mcp/tactile.py
jlongvalRobotiq: Still the plan: the driver publishes rises and exposes `tare` (#72), and this module shrinks to `contact_signal`. Nothing moved here this round.

mcp/gripper_mcp/tactile_backend.py
jlongvalRobotiq: Dropped. Nothing in the whole train reads it, including the tactile-guided close in #73, so the field was a promise without a consumer. #66 loses the `TactileSensor/Dynamic` subscription and its tests with it. If a first-touch detector wants the channel later, it comes back together with that consumer.

mcp/gripper_mcp/tactile.py
jlongvalRobotiq: `ValueError` now, and `TactileSpec.baseline_samples` is `Field(gt=0)` on #65, so the only call site cannot reach it.

mcp/gripper_mcp/tactile.py
jlongvalRobotiq: Renamed `peak_rise`; #65 still publishes it as `peak_taxel_counts`.

mcp/gripper_mcp/tactile.py
jlongvalRobotiq: One paragraph in the module docstring now: 1.0 means every taxel on both pads at the working ceiling, a firm one-sided contact reads about half, thresholds are in those units. The per-pad question continues on #65.

PR #64, opened 2026-09-04
Issue 27 38/survive a missing gripper

### LINKED ISSUE #27: robotiq_control.launch.py: spawners retry forever after ros2_control_node exits
When `ros2_control_node` exits — for example when no gripper is attached and `on_configure` fails — the three spawners launched by `robotiq_control.launch.py` keep polling `/controller_manager/list_controllers` indefinitely instead of shutting down with the rest of the launch. The user has to Ctrl-C the whole launch, and until they do the processes stay up and spam the logs.

The connection failure itself is now reported cleanly (an actionable message naming 24 V power, RS-485 wiring, `com_port` / `slave_address` / `baudrate`), so what remains is the launch never terminating.

## Reproduce

Two ways. The second needs no hardware and no container and takes seconds, so it is the
easier one for comparing before and after.

### 1. Full bringup with no gripper

**In the container**, with the gripper unplugged so `run.sh` maps no serial device:

```bash
cd <your clone of robotiq/ros>
./docker/run.sh build       # rebuild the image from the current checkout: run.sh reuses an
                            # existing image, so a branch switch alone changes nothing inside it
./docker/run.sh gripper     # logs "detecting serial devices (/dev/ttyUSB*) ..." and finds none

# then, at the container prompt:
ros2 launch robotiq_description robotiq_control.launch.py com_port:=/dev/ttyUSB0
```

*Before:* the three spawners poll `/controller_manager/list_controllers` about every 10 s
indefinitely, and the launch never ends — see Observed below.

*After:* each spawner gives up and exits non-zero. On a branch that also carries the #38 fix the
controller manager stays up, so they fail in about a second with `Unable to activate controller
... is not available` rather than timing out at all.

### 2. The spawner on its own — the quickest before/after

Run with no bringup active, so there is no controller manager to find:

```bash
# before: no absolute timeout
time ros2 run controller_manager spawner joint_state_broadcaster -c /controller_manager
```

Measured: still running at 25 s, one retry logged every ~10 s, never exits — Ctrl-C to stop.

```bash
# after: the whole change is this one flag
time ros2 run controller_manager spawner joint_state_broadcaster -c /controller_manager \
  --controller-manager-timeout 5
```

Measured: `[FATAL] Could not contact service /controller_manager/list_controllers`, exit code 1,
after ~5 s.

## Observed

```
[spawner-3] [INFO] [spawner_joint_state_broadcaster]:
  waiting for service /controller_manager/list_controllers to become available...
[spawner-4] [WARN] [spawner_robotiq_gripper_controller]:
  Could not contact service /controller_manager/list_controllers
[spawner-5] [WARN] [spawner_robotiq_activation_controller]:
  Could not contact service /controller_manager/list_controllers
... repeating every ~10 s, indefinitely
```

## Analysis

- The launch file declares no `on_exit` action tying the spawners' lifetime to `ros2_control_node`.
- `controller_manager/spawner` loops by default with no absolute timeout — only a periodic retry with a log line.

## Possible fixes

1. **In the launch file:** register an event handler on `OnProcessExit(ros2_control_node)` that emits a `Shutdown` event. A dozen lines of Python, local to this repo.
2. **Upstream in `ros2_control`:** give `spawner` a `--max-retries` / `--max-wait-time` that makes it bail after N attempts. Cleaner conceptually, but needs an upstream PR.

Option 1 is enough to make a failed bringup exit cleanly; option 2 would be a community contribution on top.

### LINKED ISSUE #38: ros2_control_node aborts if the gripper is not connected when the driver starts
If `/dev/ttyUSB0` does not exist (gripper unplugged) or is inaccessible when `robotiq_control.launch.py` starts, `ros2_control_node` dies with `SIGABRT` instead of staying `unconfigured` and retrying. The only way back is to plug the gripper in and relaunch everything.

Carried over from PickNik: [PickNikRobotics/ros2_robotiq_gripper#88](https://github.com/PickNikRobotics/ros2_robotiq_gripper/issues/88).

## Reproduce

Either path produces the same failure.

**In the container**, with the gripper unplugged so `run.sh` maps no serial device:

```bash
cd <your clone of robotiq/ros>
./docker/run.sh build       # rebuild the image from the current checkout: run.sh reuses an
                            # existing image, so a branch switch alone changes nothing inside it
./docker/run.sh gripper     # logs "detecting serial devices (/dev/ttyUSB*) ..." and finds none

# then, at the container prompt:
ros2 launch robotiq_description robotiq_control.launch.py com_port:=/dev/ttyUSB0
```

`ROS_DISTRO=humble ./docker/run.sh gripper` reproduces it on the Humble image too.

That form needs neither hardware nor a container and takes the identical path: the port
lookup fails before anything is opened.

## Observed (ROS 2 Jazzy, abridged)

```
[resource_manager]: 'configure' hardware 'RobotiqGripperHardwareInterface'
[RobotiqGripperHardwareInterface]: Cannot connect to the Robotiq gripper:
    SerialIOException: no serial port named '/dev/ttyUSB0' (is the device connected?)
[resource_manager]: Failed to 'configure' hardware 'RobotiqGripperHardwareInterface'
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed to set the initial state of the component
           : RobotiqGripperHardwareInterface to active
... (stack trace through ControllerManager::init_resource_manager)
Aborted (Signal sent by tkill() 30 0)
[ERROR] [ros2_control_node-1]: process has died [pid 30, exit code -6, ...]
```

## Analysis

1. The SDK raises `SerialIOException` when the port cannot be opened.
2. The hardware interface's `on_configure` correctly returns `ERROR` — that part follows the ros2_control lifecycle.
3. **But** the launch configuration asks for initial state `active`. When configure fails, `ControllerManager` re-throws as an uncaught `std::runtime_error` → `std::terminate` → `SIGABRT`. The whole `ros2_control_node` dies.

The diagnostic half of this has already improved: the message now names the port and asks whether the device is connected, instead of surfacing a raw `IO Exception (2)` with a library source path. The crash itself is unchanged.

## Expected

Configure fails → the hardware stays `unconfigured` → the node keeps running, retries periodically (or exposes a service for a manual retry). The user can plug the gripper in afterwards and recover without relaunching.

## Related

After this crash the three spawners poll `/controller_manager/list_controllers` forever — #27.

ebarnett3:
## Survive a missing gripper at bringup

Closes #27. Closes #38 on Jazzy and newer; Humble is discussed below.

Launching `robotiq_control.launch.py` with no gripper attached used to fail
twice over. `ros2_control_node` asks for the hardware component to come up
`active`, so when `on_configure` cannot open the port the controller manager
rethrows and the process aborts (#38). Where the abort completes, the three
spawners then poll `/controller_manager/list_controllers` every 10 s forever
(#27); on some hosts the abort does not complete and the process wedges in its
backtrace handler, ignoring `SIGINT` and `SIGTERM`. Either way the only way
forward was Ctrl-C and a relaunch with the gripper plugged in.

Four commits, one per change:

**Bound the spawners' wait.** Each spawner gets `--controller-manager-timeout 15`,
so a controller manager that never answers makes them exit non-zero instead of
retrying indefinitely. The value is the driver's default `activation_timeout`:
the manager only answers once the hardware is active, so a gripper still
recovering from a fault must not read as missing. An unplugged gripper fails in
about a second regardless, so the bound only delays the verdict where the
manager never comes at all.
An `OnProcessExit` handler on `ros2_control_node` also ends the launch when the
node does exit. `shutdown_on_failure:=false` opts out of that for includers;
`gripper_tactile_viz.launch.py` passes it, so the tactile stack keeps streaming
without a gripper, as it did before.

**Keep the node alive.** `shutdown_on_initial_state_failure: false` in the
Jazzy controller config. A failed connect leaves the component `unconfigured`
and the node running, so the spawners fail fast and the user can recover
without relaunching.

**Say how to retry the connection.** The driver's connect error now ends with
the retry for the hardware component, since it is the one place that knows the
component's name:

```
Cannot connect to the Robotiq gripper on /dev/ttyUSB0 ...: <cause>. Once it is
connected, retry with: ros2 control set_hardware_component_state RobotiqGripperHardwareInterface active
```

**Say why the bringup failed, and the way back.** After the last spawner has
exited, and only if one failed, the launch prints both recovery steps in order.
Bringing the hardware back leaves the three controllers loaded but inactive
(nothing in the controller manager reactivates them), so the second step is a
spawner re-run, built from the spawners' own argument list so the two cannot
drift, which configures and activates what is already loaded:

```
ros2 control set_hardware_component_state RobotiqGripperHardwareInterface active
ros2 run controller_manager spawner joint_state_broadcaster robotiq_gripper_controller robotiq_activation_controller --controller-manager /controller_manager --controller-manager-timeout 15 --param-file /tmp/launch_params_...
```

The component names come from the description the launch loads, so a custom
`model:=` gets its own names.

The hint stays silent when the spawners were killed rather than failed (Ctrl-C,
or the launch's own shutdown) and under `use_fake_hardware` or `sim_topic_based`,
where no serial link can be the cause.

## Humble

Humble's `controller_manager` has no `shutdown_on_initial_state_failure`
parameter, so its config is unchanged and the node still aborts or wedges
there. When it aborts, the node-exit handler says to reconnect and relaunch and
ends the launch. When it wedges there is no exit to react to, so the spawners'
timeout is what ends the bringup: once they give up the launch prints the same
line and ends itself (unless `shutdown_on_failure:=false`), which is what gets
a wedged node killed.

## Retrying on its own

#38 also asked for the node to retry the connection by itself. This PR leaves
that to the user on purpose: reconnecting with backoff is a general grippers
need, so the retry loop belongs in the grippers SDK, which plans it. #71 tracks
the ROS side once that lands.

## Testing

Verified on Jazzy against a 2F-85, on the host, with the port held busy by
another process to stand in for an unplugged gripper:

| | gripper unplugged | plugged in after launch |
|---|---|---|
| before | node aborts (or wedges); spawners retry forever | relaunch only |
| after | node stays up; spawners exit in ~1 s; both hints printed | `set_hardware_component_state ... active` alone leaves all three controllers inactive; the spawner re-run brings them `active` and `/joint_states` flows |

Ctrl-C two seconds into a bringup prints no hint. The hint above is the one printed on the bench, param file path included. With the gripper attached
from the start the bringup is unchanged.

Tests drive the launch's exit handlers per spawner and check: the hint fires
once, after the last exit, for `[1,1,1]`, `[0,1,0]` and `[1,0,0]`; nothing
fires for success, for signal exits, once shutdown is under way, or under the
mock and simulated plugins; Humble gets the relaunch wording plus a `Shutdown`,
suppressed by `shutdown_on_failure:=false`; the control-node exit handler
emits `Shutdown` and is absent when opted out; the spawners carry the 10 s
timeout. The `controller_manager` blocks are compared exhaustively against the
settings each config may carry, so a misspelled setting cannot slip through.

### DIFF
--- a/grippers/robotiq_description/config/robotiq_controllers.yaml
+++ b/grippers/robotiq_description/config/robotiq_controllers.yaml
@@ -4,6 +4,12 @@
 controller_manager:
   ros__parameters:
     update_rate: 500  # Hz
+    # Jazzy defaults this to true and takes ros2_control_node down when the
+    # gripper is not connected; Lyrical defaults to false. Set explicitly so every
+    # distro keeps the node up and the connection can be retried from the CLI.
+    # Humble has no such parameter.
+    hardware_components_initial_state:
+      shutdown_on_initial_state_failure: false
     joint_state_broadcaster:
       type: joint_state_broadcaster/JointStateBroadcaster
     # Jazzy and newer: parallel_gripper_action_controller replaces gripper_controllers package
--- a/grippers/robotiq_description/launch/robotiq_control.launch.py
+++ b/grippers/robotiq_description/launch/robotiq_control.launch.py
@@ -288,8 +288,9 @@ def generate_launch_description():
         launch.actions.DeclareLaunchArgument(
             name="shutdown_on_failure",
             default_value="true",
-            description="End the launch when ros2_control_node exits. Set false when "
-            "including this file next to nodes that should outlive the gripper",
+            description="End the launch when ros2_control_node exits, or, on Humble, when no "
+            "controller could be activated. Set false when including this file next to "
+            "nodes that should outlive the gripper",
         )
     )
 
@@ -301,7 +302,8 @@ def generate_launch_description():
         )
     }
 
-    controllers_file = ControllersFile(os.environ.get("ROS_DISTRO"), topic_based)
+    distro = os.environ.get("ROS_DISTRO")
+    controllers_file = ControllersFile(distro, topic_based)
     initial_joint_controllers = ParameterFile(
         PathJoinSubstitution([description_pkg_share, "config", controllers_file]),
         allow_substs=True,
@@ -354,12 +356,66 @@ def generate_launch_description():
             condition=condition,
         )
 
-    joint_state_broadcaster_spawner = spawner("joint_state_broadcaster")
-    robotiq_gripper_controller_spawner = spawner("robotiq_gripper_controller")
-    # The reactivate_gripper GPIO this controller claims is declared for the
-    # driver and the mock only; the topic_based plugin has nothing to reactivate.
-    robotiq_activation_controller_spawner = spawner(
-        "robotiq_activation_controller", condition=UnlessCondition(topic_based)
+    # The reactivate_gripper GPIO the activation controller claims is declared for
+    # the driver and the mock only; the topic_based plugin has nothing to reactivate.
+    spawned_controllers = {
+        "joint_state_broadcaster": None,
+        "robotiq_gripper_controller": None,
+        "robotiq_activation_controller": UnlessCondition(topic_based),
+    }
+    spawners = [
+        spawner(name, condition) for name, condition in spawned_controllers.items()
+    ]
+
+    def starts(action, context):
+        return action.condition is None or action.condition.evaluate(context)
+
+    def is_set(context, name):
+        return evaluate_condition_expression(context, [LaunchConfiguration(name)])
+
+    # After a failed bringup the last line on screen is otherwise a spawner's
+    # "process has died", which says nothing about the cause or the way out.
+    returncodes = []
+
+    def hint_after_the_last_spawner_exits(event, context):
+        returncodes.append(event.returncode)
+        if context.is_shutdown or len(returncodes) < sum(
+            starts(s, context) for s in spawners
+        ):
+            return None
+        # A negative code is a signal: the spawners were killed, they did not fail.
+        if not any(code > 0 for code in returncodes):
+            return None
+        # Only a real gripper can be unplugged.
+        if any(is_set(context, flag) for flag in HARDWARE_FLAGS):
+            return None
+        if distro == "humble":
+            hint = launch.actions.LogInfo(
+                msg="A controller could not be activated. If the gripper is not "
+                "connected, ros2_control_node did not survive it (see its error above); "
+                "reconnect the gripper and relaunch."
+            )
+            if not is_set(context, "shutdown_on_failure"):
+                return [hint]
+            return [hint, launch.actions.Shutdown(reason="no controller activated")]
+        names = " ".join(
+            name
+            for name, action in zip(spawned_controllers, spawners)
+            if starts(action, context)
+        )
+        return [
+            launch.actions.LogInfo(
+                msg="A controller could not be activated. If the gripper is not "
+                "connected, follow the driver's error above to bring the hardware "
+                f"back, then re-run the spawners: ros2 run controller_manager spawner {names}"
+            )
+        ]
+
+    spawner_hint = launch.actions.RegisterEventHandler(
+        launch.event_handlers.OnProcessExit(
+            target_action=lambda action: action in spawners,
+            on_exit=hint_after_the_last_spawner_exits,
+        )
     )
 
     shutdown_on_control_node_exit = launch.actions.RegisterEventHandler(
@@ -374,11 +430,10 @@ def generate_launch_description():
         OpaqueFunction(function=reject_conflicting_hardware_flags),
         control_node,
         robot_state_publisher_node,
-        joint_state_broadcaster_spawner,
-        robotiq_gripper_controller_spawner,
-        robotiq_activation_controller_spawner,
+        *spawners,
         rviz_node,
         shutdown_on_control_node_exit,
+        spawner_hint,
     ]
 
     return launch.LaunchDescription(
--- a/grippers/robotiq_description/launch/robotiq_control.launch.py
+++ b/grippers/robotiq_description/launch/robotiq_control.launch.py
@@ -41,9 +41,11 @@ from launch.conditions import (
     UnlessCondition,
     evaluate_condition_expression,
 )
+from launch.utilities import normalize_to_list_of_substitutions, perform_substitutions
 import launch_ros
 from launch_ros.parameter_descriptions import ParameterFile
 import os
+import re
 
 # Humble has no parallel_gripper_controller package, so it needs its own
 # controller config, and the topic_based plugin exports a different
@@ -137,6 +139,15 @@ class ParameterFilePath(Substitution):
 
     def perform(self, context):
         return str(self.parameter_file.evaluate(context))
+
+
+CONTROLLER_MANAGER_TIMEOUT = "15"
+
+
+def hardware_component_names(context):
+    """Names of the <ros2_control> components in the description the launch loads."""
+    urdf = perform_substitutions(context, [xacro_command()])
+    return re.findall(r'<ros2_control\s+name="([^"]+)"', urdf)
 
 
 # Joint carrying the position command, per gripper_model. The launch argument
@@ -340,19 +351,22 @@ def generate_launch_description():
     # then dies with "parameter 'joint' is not initialized". --param-file has been
     # a spawner option since Humble, and each node reads only its own section of
     # the file, so this is correct on every supported distro.
+    def spawner_arguments(*controller_names):
+        return [
+            *controller_names,
+            "--controller-manager",
+            "/controller_manager",
+            "--controller-manager-timeout",
+            CONTROLLER_MANAGER_TIMEOUT,
+            "--param-file",
+            ParameterFilePath(initial_joint_controllers),
+        ]
+
     def spawner(controller_name, condition=None):
         return launch_ros.actions.Node(
             package="controller_manager",
             executable="spawner",
-            arguments=[
-                controller_name,
-                "--controller-manager",
-                "/controller_manager",
-                "--controller-manager-timeout",
-                "10",
-                "--param-file",
-                ParameterFilePath(initial_joint_controllers),
-            ],
+            arguments=spawner_arguments(controller_name),
             condition=condition,
         )
 
@@ -373,11 +387,40 @@ def generate_launch_description():
     def is_set(context, name):
         return evaluate_condition_expression(context, [LaunchConfiguration(name)])
 
+    def uses_real_gripper(context):
+        return not any(is_set(context, flag) for flag in HARDWARE_FLAGS)
+
+    relaunch_hint = launch.actions.LogInfo(
+        msg="The bringup failed. If the error above says the gripper could not be "
+        "connected, connect it and relaunch: on Humble, ros2_control_node cannot "
+        "recover from a failed connection without a restart."
+    )
+
+    def recovery_hint(context):
+        """Both halves of the recovery, in order: the hardware, then the controllers.
+
+        Bringing the component back leaves the controllers loaded but inactive;
+        re-running the spawners is what configures and activates them again.
+        """
+        names = [n for n, s in zip(spawned_controllers, spawners) if starts(s, context)]
+        arguments = " ".join(
+            perform_substitutions(context, normalize_to_list_of_substitutions(a))
+            for a in spawner_arguments(*names)
+        )
+        steps = [
+            f"ros2 control set_hardware_component_state {component} active"
+            for component in hardware_component_names(context)
+        ] + [f"ros2 run controller_manager spawner {arguments}"]
+        return launch.actions.LogInfo(
+            msg="A controller could not be activated. If the gripper is not connected, "
+            "reconnect it, then run, in order:\n  " + "\n  ".join(steps)
+        )
+
     # After a failed bringup the last line on screen is otherwise a spawner's
     # "process has died", which says nothing about the cause or the way out.
     returncodes = []
 
-    def hint_after_the_last_spawner_exits(event, context):
+    def on_spawner_exit(event, context):
         returncodes.append(event.returncode)
         if context.is_shutdown or len(returncodes) < sum(
             starts(s, context) for s in spawners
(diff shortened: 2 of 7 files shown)


grippers/robotiq_description/launch/robotiq_control.launch.py
```diff
@@ -292,28 +302,92 @@ def spawner(controller_name, condition=None):
...
+        names = " ".join(
+            name
+            for name, action in zip(spawned_controllers, spawners)
+            if starts(action, context)
+        )
+        return [
+            launch.actions.LogInfo(
+                msg="A controller could not be activated. If the gripper is not "
+                "connected, follow the driver's error above to bring the hardware "
+                f"back, then re-run the spawners: ros2 run controller_manager spawner {names}"
```
jlongvalRobotiq: This retry command is missing the `--param-file` command i think ? 
Also, 
Could the hint be built from the same argument list `spawner()` uses? Something like extracting the args into a helper both `spawner()` and the hint call, so they can't drift. Would also carry `--controller-manager-timeout` for free.

grippers/robotiq_description/launch/robotiq_control.launch.py
```diff
@@ -292,28 +302,92 @@ def spawner(controller_name, condition=None):
...
+            starts(s, context) for s in spawners
+        ):
+            return None
+        # A negative code is a signal: the spawners were killed, they did not fail.
+        if not any(code > 0 for code in returncodes):
+            return None
+        # Only a real gripper can be unplugged.
+        if any(is_set(context, flag) for flag in HARDWARE_FLAGS):
+            return None
+        if distro == "humble":
```
jlongvalRobotiq, line 433: if the node aborts instead (wich is mentionned as a possibility in #38), `shutdown_on_control_node_exit` fires `Shutdown()` right away and the `context.is_shutdown` statement above makes it so we wont have a hint. 

Options: print a one-line hint from the control-node exit handler itself when `distro == "humble"`, or let this handler still emit the hint when shutdown was triggered by the control node exiting. Either way a test for "Humble + node exits before spawners" would pin it.

grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -227,9 +227,11 @@ rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn Roboti
    catch(const std::exception& e)
    {
       RCLCPP_ERROR(kLogger,
-                   "Cannot connect to the Robotiq gripper on %s: %s",
+                   "Cannot connect to the Robotiq gripper on %s: %s. Once it is connected, retry with: ros2 control "
+                   "set_hardware_component_state %s active",
```
jlongvalRobotiq: Recovery is now two steps in two places: this message says `set_hardware_component_state … active`, the launch hint (L360-366) says re-run the spawners. Neither mentions the other, and the launch hint is what the user reads last. If they run only that, the spawners fail again because the component is still inactive.

Suggestion: the launch hint prints both commands in order (it already has names; the component name is in the URDF the launch loads), or at minimum says "run the driver's command first, then …". The driver message could then stay as-is.

grippers/robotiq_description/launch/robotiq_control.launch.py
```diff
@@ -292,28 +302,92 @@ def spawner(controller_name, condition=None):
                 controller_name,
                 "--controller-manager",
                 "/controller_manager",
+                "--controller-manager-timeout",
+                "10",
```
jlongvalRobotiq: The `"10"` deserves a name and a short why. It's also a bit tight: the controller manager only answers after the serial port is open, and on a slow machine with a real gripper plugged in, 10 s might not be enough. We'd then print "not connected" when it is. 15-30 s costs nothing on the unplugged path.

grippers/robotiq_description/launch/robotiq_control.launch.py
```diff
@@ -292,28 +302,92 @@ def spawner(controller_name, condition=None):
...
+        return action.condition is None or action.condition.evaluate(context)
+
+    def is_set(context, name):
+        return evaluate_condition_expression(context, [LaunchConfiguration(name)])
+
+    # After a failed bringup the last line on screen is otherwise a spawner's
+    # "process has died", which says nothing about the cause or the way out.
+    returncodes = []
+
+    def hint_after_the_last_spawner_exits(event, context):
```
jlongvalRobotiq: Nit: the name says "hint", but the first line does the bookkeeping. `on_spawner_exit` fits better.

grippers/robotiq_description/test/test_controller_config.py
```diff
@@ -372,3 +405,173 @@ def test_controller_names_and_update_rate_are_identical_across_distros(config):
...
+    # SIGINT from a Ctrl-C or from the launch's own Shutdown is not a failure.
+    assert drive_spawner_exits(monkeypatch, [-2, -2, -2]) == [[], [], []]
+
+
+@requires_launch
+def test_launch_stays_quiet_once_shutting_down(monkeypatch):
+    from launch.events.process import ProcessExited
+
+    entities, context = launch_entities("jazzy", monkeypatch)
+    context._set_is_shutdown(True)
```
jlongvalRobotiq, line 588: Nit: `_set_is_shutdown` is a private `launch` function, and `dispatch()` calls the handlers by hand. Works now, but a `launch` update could break these tests. Not blocking, just noting.

grippers/robotiq_description/config/robotiq_controllers.yaml
```diff
@@ -4,6 +4,12 @@
 controller_manager:
   ros__parameters:
     update_rate: 500  # Hz
+    # Jazzy defaults this to true and takes ros2_control_node down when the
+    # gripper is not connected; Lyrical defaults to false. Set explicitly so every
+    # distro keeps the node up and the connection can be retried from the CLI.
+    # Humble has no such parameter.
+    hardware_components_initial_state:
+      shutdown_on_initial_state_failure: false
```
jlongvalRobotiq, line 12: Not blocking: #38 asked for the node to retry on its own. This PR stays alive but leaves the retry to the user. That's a fair call — just say so in the PR text and open a follow-up or keep #38 open.

grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -227,9 +227,11 @@ rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn Roboti
...
                    describeLink(parameters_.connection).c_str(),
-                   e.what());
+                   e.what(),
+                   info_.name.c_str());
```
jlongvalRobotiq, line 234: Nit: `info_.name` isn't quoted in the printed command. Would break if the name had a space.

jlongvalRobotiq (commented):
A few smaller notes to go with the three inline comments above.

Nit: CONTRIBUTING says the only Humble/Jazzy difference is the controller type. This PR adds another one (what happens when the gripper is unplugged at bringup). Worth one line there or in the README's supported-distros section.

jlongvalRobotiq (approved):

grippers/robotiq_description/launch/robotiq_control.launch.py
```diff
@@ -292,28 +302,92 @@ def spawner(controller_name, condition=None):
                 controller_name,
                 "--controller-manager",
                 "/controller_manager",
+                "--controller-manager-timeout",
```
mbegin-robotiq: This 10 s spawner timeout is shorter than the driver's own 15 s `activation_timeout`. `controller_manager` only advertises its services once the resource manager has activated the hardware, so a gripper that is present but recovering from a fault will now fail the bringup, where the spawners previously waited indefinitely. Suggest raising this to at least the activation timeout (or deriving it from the same value).

grippers/robotiq_description/launch/robotiq_control.launch.py
```diff
@@ -292,28 +302,92 @@ def spawner(controller_name, condition=None):
...
+        returncodes.append(event.returncode)
+        if context.is_shutdown or len(returncodes) < sum(
+            starts(s, context) for s in spawners
+        ):
+            return None
+        # A negative code is a signal: the spawners were killed, they did not fail.
+        if not any(code > 0 for code in returncodes):
+            return None
+        # Only a real gripper can be unplugged.
+        if any(is_set(context, flag) for flag in HARDWARE_FLAGS):
```
mbegin-robotiq: This `HARDWARE_FLAGS` early return runs before the `distro == "humble"` branch, so on Humble with `use_fake_hardware` / `sim_isaac` a failed bringup emits neither the hint nor the `Shutdown` — which contradicts the documented `shutdown_on_failure` behaviour. Moving the Humble branch above this check (or letting the shutdown path run regardless of the flags) would keep the two consistent.

grippers/robotiq_description/launch/robotiq_control.launch.py
mbegin-robotiq: The printed re-run command omits `--param-file`, which the comment further up in this file says is required from Lyrical on. If the first spawn died before the controller parameters were set, following this hint fails with `parameter 'joint' is not initialized`. Worth appending the same `--param-file` argument the spawners above use.

mbegin-robotiq (approved):
Approach looks good — the bringup now survives a missing gripper instead of taking the whole launch down. Approving; three things worth fixing before merge, left inline.

Non-blocking: the README isn't updated for the new `shutdown_on_failure` arg, nor for the fact that on Jazzy a missing gripper now leaves the launch running instead of exiting.

grippers/robotiq_description/launch/robotiq_control.launch.py
ebarnett3: Done in fd3bd8a: `spawner_arguments()` now feeds both `spawner()` and the hint, so the re-run carries `--controller-manager`, `--controller-manager-timeout` and `--param-file` (resolved to the same temp file the first spawn used). On the bench the re-run worked without `--param-file` only because the controllers were already loaded; with it the hint also covers the never-loaded case.

grippers/robotiq_description/launch/robotiq_control.launch.py
ebarnett3: Right, the abort path lost the hint. In fd3bd8a the control-node exit handler prints the relaunch line itself on Humble (and still emits `Shutdown` unless `shutdown_on_failure:=false`). Pinned by `test_launch_hints_when_the_node_aborts_on_humble`.

grippers/robotiq_driver/src/hardware_interface.cpp
ebarnett3: Agreed. The launch hint now prints both steps in order (fd3bd8a): `set_hardware_component_state <name> active` for every `<ros2_control>` component in the description the launch actually loads, then the spawner re-run. The driver message stays as is.

grippers/robotiq_description/launch/robotiq_control.launch.py
ebarnett3: Named `CONTROLLER_MANAGER_TIMEOUT` and set to 15 s (fd3bd8a), the low end of your range: it matches the driver's default `activation_timeout`, which is what bounds a healthy-but-recovering gripper. We kept it short on purpose, to fail fast; the bound only matters where the manager never answers, since an unplugged gripper on Jazzy fails in about a second regardless.

grippers/robotiq_description/launch/robotiq_control.launch.py
ebarnett3: Renamed `on_spawner_exit` (fd3bd8a).

grippers/robotiq_description/test/test_controller_config.py
ebarnett3: Left as is. `LaunchContext` has no public setter for the shutdown flag, and the alternative is driving a real `LaunchService` with fake processes, which is a bigger harness than this launch file deserves. If a `launch` update breaks it, the failure will be loud.

grippers/robotiq_description/config/robotiq_controllers.yaml
ebarnett3: Closing #38 here and opened #71 for reconnect-with-backoff. The retry loop is a general grippers need, so it goes in the grippers SDK; #71 tracks the ROS side once that lands. The PR text now says so.

grippers/robotiq_driver/src/hardware_interface.cpp
ebarnett3: Quoted (fd3bd8a).

grippers/robotiq_description/launch/robotiq_control.launch.py
ebarnett3: Raised to 15 s, the driver's default `activation_timeout`, as a named constant (fd3bd8a). Not derived from the driver value itself: it is a hardware parameter in the xacro, which the launch does not see.

grippers/robotiq_description/launch/robotiq_control.launch.py
ebarnett3: Good catch. In fd3bd8a the Humble branch runs regardless of the hardware flags: only the reconnect wording is hardware-only, `Shutdown` follows `shutdown_on_failure` under the mock and the topic-based plugin too. Pinned by `test_launch_still_ends_on_humble_without_a_gripper`.

grippers/robotiq_description/launch/robotiq_control.launch.py
ebarnett3: Appended (fd3bd8a): the hint is built from the spawners' own argument list, so it carries `--param-file` pointing at the temp file the launch still holds, plus the manager and timeout flags.

ebarnett3:
Review notes addressed in b6084a5: CONTRIBUTING now names the missing-gripper behaviour as the second Humble/Jazzy difference, and the README supported-distros section describes it per distro along with `shutdown_on_failure`.

PR #65, opened 2026-09-04
feat(mcp): tactile datasheet, TactileService and read/tare/verify_grasp tools

jlongvalRobotiq:
## Change

Ninth PR of the `mcp/` series (stacked on #62, now merged). Tactile reaches the tool surface; the ROS source and `grasp_until_contact` are the next PRs.

1. **Datasheets + wiring** — the TSF-85 pads are a separate product fitted onto a 2F gripper, so they get their own sheet, `datasheets/tactile/robotiq_tsf_85.yaml` (the 7x4 two-pad layout `robotiq_tsf` publishes, `full_scale_counts`, `contact_threshold` as a floor, `noise_margin`, `baseline_samples`); `robotiq_2f_85.yaml` names it with `tactile_model` and the loader stitches the two so consumers keep reading `spec.tactile`. The 2F-140 names no pads. A `grippers.yaml` entry opts in with `tactile: ros`, and startup fails clearly if the model names no tactile pads. `gripper_list_grippers` now reports each gripper's tactile source (or `null`).
2. **`tactile_service.py`** — a `TactileService` next to `GripperService` rather than inside it: tactile is optional per gripper and owns its own state, the rest baseline. The baseline is taken on demand from an open gripper, retaken by `gripper_tare_tactile`, and refused with a pointer to the tare when the fingers are not open. A tare averages `baseline_samples` *distinct* frames: the tactile protocol gained `sample(count)` (first commit), so a source that caches its latest message cannot hand back the same frame a thousand times. The tare also measures its own noise: the threshold in force is `max(contact_threshold, noise_margin x the peak rest signal the tare saw)`, reported by the tare and echoed on every reading, so real pads whose noise sits above the twin-tuned floor (Eric measured 0.035 against a 0.02 floor) still get a threshold above their own noise. `verify_grasp` reads the gripper state once and derives both the opening it reports and the fingers-met test from that single reading, so the verdict and the opening describe the same instant. `judge_grasp`: contact with the fingers apart is `held`; fingers that met are `closed_on_nothing` whatever the pads say (TSF-85 pads touch each other when fully closed); fingers apart with quiet pads is `no_contact`, the slipped-or-stopped-outside-the-pads case.
3. **Three tools** — `gripper_read_tactile` (signal, per-pad split as each pad's fraction of its own ceiling, hottest taxel, contact flag), `gripper_tare_tactile` (samples, rest mean, `noise_floor`, `threshold`), `gripper_verify_grasp`. Tool registration is now two helpers (`register_gripper_tools` / `register_tactile_tools`) and `build_service` became `build_services(wiring, make_backend=..., make_tactile=...)`, returning both. The tactile factory is a parameter like the backend's: the default refuses a `tactile` entry by name until the ROS source lands, the tests hand in the fake pads.
The split into two services is also what keeps `GripperService` inside the coupling budget of the maintainability gate this repo's `mcp/` code is held to; the first cut, with tactile folded in, tripped it.

## Verification

- `uv run pytest`: 182 passed (34 new): `sample(count)` on the mock; `rest_noise`; `noise_margin` validation; quiet pads keep the datasheet floor; pads scripted at Eric's bench numbers raise the threshold to ~0.07, an empty open gripper at 0.035 is `no_contact`, an external hold at 0.8 is `held`; a verification reads the gripper state once; the 2F-85 names the TSF-85 sheet / 2F-140 names none / every shipped tactile sheet is named by a gripper / a tactile block inlined in a gripper sheet is rejected / a missing tactile sheet is named / `tactile: mock` is not a wiring option; tare, read at rest, grasp-on-cube presses both pads, verify → `held` / `no_contact` / `closed_on_nothing`, stall-outside-the-pads is not a hold, first read self-baselines from an open gripper, closed-with-no-baseline asks for a tare, gripper without pads refuses, unknown gripper still `UnknownGripperError`, `judge_grasp` table; wire-level grasp→verify says `held`, tactile on a 2F-140 fails at startup, the default factory names the missing ROS tactile source, annotations.

## Follow-up
#77: measure the constants on real TSF pads (assigned to me; none on hand today).

## Stack
#62 is merged; this PR shows its own 3 commits.

### DIFF
--- a/mcp/gripper_mcp/datasheets/robotiq_2f_85.yaml
+++ b/mcp/gripper_mcp/datasheets/robotiq_2f_85.yaml
@@ -31,3 +31,8 @@ defaults:
   # Requested grip force when a call does not name one; must sit in grip_force.
   max_effort_n: 50.0
   motion_timeout_s: 10.0
+
+# Tactile pads that fit this gripper: datasheets/tactile/<tactile_model>.yaml.
+# Omit for a model with no tactile option. Whether a given gripper wears them
+# is the wiring's business (`tactile: ros` in grippers.yaml).
+tactile_model: robotiq_tsf_85
--- a/mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
+++ b/mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
@@ -17,9 +17,16 @@ pads: [left, right]
 # Unverified against a physical TSF-85; re-measure before trusting it there.
 full_scale_counts: 6160.0
 
-# Contact is declared at or above this signal. Deliberately just above the
-# noise floor: the point is to notice FIRST touch, not a firm grip.
+# Floor for the contact threshold. Deliberately just above the twin's noise
+# floor: the point is to notice FIRST touch, not a firm grip.
 contact_threshold: 0.02
+
+# The threshold used at runtime is max(contact_threshold, noise_margin x the
+# peak rest signal seen during the tare), so real pads whose noise sits above
+# the twin-tuned floor still get a threshold above their own noise. Measured
+# on a physical TSF-85: rest noise peaks near 0.035 of full scale, an external
+# hold reads about 0.8, so 2x the noise separates the two with room.
+noise_margin: 2.0
 
 # Matches the tactile SDK's findBaseline(), which averages 1000 samples.
 baseline_samples: 1000
--- a/mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
+++ b/mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
@@ -1,0 +1,25 @@
+# Robotiq TSF-85 tactile fingers datasheet. The pads are a separate product
+# fitted onto a 2F-85 and read by their own driver (robotiq_tsf), so they get
+# their own sheet; the gripper's datasheet points here with `tactile_model`.
+# Whether a given gripper has them or not is the wiring's business
+# (`tactile: ros` in grippers.yaml).
+
+model: robotiq_tsf_85
+
+# The 7x4 grid per pad is what robotiq_tsf publishes (Taxels: uint16[28]).
+rows: 7
+cols: 4
+pads: [left, right]
+
+# Signal 1.0 means every taxel on both pads is at the working ceiling:
+# 56 taxels x 110 counts. Tuned against the Isaac Sim twin's tactile
+# extension (hottest taxel ~106, no-contact floor ~0.0003 of full scale).
+# Unverified against a physical TSF-85; re-measure before trusting it there.
+full_scale_counts: 6160.0
+
+# Contact is declared at or above this signal. Deliberately just above the
+# noise floor: the point is to notice FIRST touch, not a firm grip.
+contact_threshold: 0.02
+
+# Matches the tactile SDK's findBaseline(), which averages 1000 samples.
+baseline_samples: 1000
--- a/mcp/gripper_mcp/tactile_service.py
+++ b/mcp/gripper_mcp/tactile_service.py
@@ -1,0 +1,201 @@
+"""Tactile tools over a gripper's TSF-85 pads.
+
+Kept apart from `GripperService` because tactile is optional per gripper and
+has its own state, the rest baseline. The one piece of domain judgement here is
+`judge_grasp`: contact on the pads with the fingers apart is a hold; fingers
+that met hold nothing whatever the pads say (on a TSF-85 the pads touch each
+other when fully closed, and read well above the contact threshold, so position
+has to make that call, never the pads).
+
+An internal hold, the fingers pressing outward on the inside of a bore, is
+recognised only weakly: the pads face away from the part, and what they read
+comes from finger deflection or the bore edge, a few times the noise floor
+against an order of magnitude for an external hold. Do not tune the threshold
+against it. The gripper's own object detection names that case outright and
+will replace the inference once the backend reads it.
+
+Per-pad signals are each pad's fraction of its own ceiling
+(`full_scale_counts / pads`), so one saturated pad reads 1.0 on its own line
+while `contact_signal` stays normalised across both.
+
+The contact threshold is set at tare time, not copied from the datasheet: the
+rest frames the tare averages also show how noisy the pads are, and the
+threshold is the datasheet floor or `noise_margin` times that measured noise,
+whichever is higher. Real pads are far noisier than the twin the floor was
+tuned on, and a threshold under the noise makes an empty open gripper read as
+a hold.
+"""
+
+import threading
+from dataclasses import dataclass
+
+from gripper_mcp.config import TactileSpec
+from gripper_mcp.models import (
+    GraspVerdict,
+    GraspVerification,
+    TactileReadingResult,
+    TactileTareResult,
+)
+from gripper_mcp.service import GripperService, timestamp
+from gripper_mcp.tactile import (
+    TactileBaseline,
+    average_readings,
+    contact_signal,
+    pad_sums,
+    peak_rise,
+    rest_noise,
+)
+from gripper_mcp.tactile_backend import TactileBackend, TactileReading
+
+VERDICT_DETAIL: dict[GraspVerdict, str] = {
+    "held": "The pads register contact and the fingers stopped before meeting.",
+    "closed_on_nothing": "The fingers are fully closed; nothing is between them.",
+    "no_contact": "The fingers are apart but the pads register no contact.",
+}
+
+
+class TactileUnavailableError(Exception):
+    pass
+
+
+@dataclass(frozen=True)
+class Tare:
+    baseline: TactileBaseline
+    noise_floor: float
+    threshold: float
+
+
+class TactileService:
+    def __init__(
+        self,
+        grippers: GripperService,
+        pads: dict[str, tuple[TactileBackend, TactileSpec]],
+    ) -> None:
+        self._grippers = grippers
+        self._pads = pads
+        self._tares: dict[str, Tare] = {}
+        self._tare_lock = threading.Lock()
+
+    def tare(self, gripper_name: str) -> TactileTareResult:
+        tactile, spec = self._pads_for(gripper_name)
+        tare = self._capture_tare(gripper_name, tactile, spec)
+
+        return TactileTareResult(
+            gripper_name=gripper_name,
+            samples=spec.baseline_samples,
+            rest_counts_mean=round(mean_counts(tare.baseline), 2),
+            noise_floor=round(tare.noise_floor, 5),
+            threshold=round(tare.threshold, 5),
+            tactile_backend=tactile.name,
+            measured_at=timestamp(),
+        )
+
+    def read(self, gripper_name: str) -> TactileReadingResult:
+        opening_mm = self._grippers.get_state(gripper_name).opening_mm
+
+        return self._read_at(gripper_name, opening_mm)
+
+    def verify_grasp(self, gripper_name: str) -> GraspVerification:
+        opening_mm = self._grippers.get_state(gripper_name).opening_mm
+        reading = self._read_at(gripper_name, opening_mm)
+        stroke = self._grippers.stroke_of(gripper_name)
+        verdict = judge_grasp(reading.contact, stroke.fingers_met(opening_mm))
+
+        return GraspVerification(
+            gripper_name=gripper_name,
+            verdict=verdict,
+            object_held=verdict == "held",
+            opening_mm=opening_mm,
+            contact_signal=reading.contact_signal,
+            threshold=reading.threshold,
+            detail=VERDICT_DETAIL[verdict],
+            tactile_backend=reading.tactile_backend,
+        )
+
+    def _read_at(self, gripper_name: str, opening_mm: float) -> TactileReadingResult:
+        tactile, spec = self._pads_for(gripper_name)
+        tare = self._tare(gripper_name, tactile, spec, opening_mm)
+
+        return self._reading(gripper_name, tactile, tare, spec)
+
+    def _pads_for(self, gripper_name: str) -> tuple[TactileBackend, TactileSpec]:
+        self._grippers.assert_known(gripper_name)
+        if gripper_name not in self._pads:
+            raise TactileUnavailableError(
+                f"Gripper '{gripper_name}' has no tactile pads configured."
+            )
+        return self._pads[gripper_name]
+
+    def _tare(
+        self,
+        gripper_name: str,
+        tactile: TactileBackend,
+        spec: TactileSpec,
+        opening_mm: float,
+    ) -> Tare:
+        if gripper_name in self._tares:
+            return self._tares[gripper_name]
+        if not self._grippers.stroke_of(gripper_name).fingers_open(opening_mm):
+            raise TactileUnavailableError(
+                "No tactile baseline yet, and the fingers are not fully open to "
+                "capture one. Open the gripper, or call gripper_tare_tactile."
+            )
+        return self._capture_tare(gripper_name, tactile, spec)
+
+    def _capture_tare(
+        self, gripper_name: str, tactile: TactileBackend, spec: TactileSpec
+    ) -> Tare:
+        with self._tare_lock:
+            tare = measure_tare(tactile.sample(spec.baseline_samples), spec)
+            self._tares[gripper_name] = tare
+        return tare
+
+    def _reading(
+        self,
+        gripper_name: str,
+        tactile: TactileBackend,
+        tare: Tare,
+        spec: TactileSpec,
+    ) -> TactileReadingResult:
+        reading = tactile.read_tactile()
+        signal = contact_signal(reading, tare.baseline, spec.full_scale_counts)
+        per_pad = zip(reading.pads, pad_sums(reading, tare.baseline))
+        pad_full_scale = spec.full_scale_counts / len(reading.pads)
+
+        return TactileReadingResult(
+            gripper_name=gripper_name,
+            contact=signal >= tare.threshold,
+            contact_signal=round(signal, 5),
+            threshold=round(tare.threshold, 5),
+            pad_signals={
+                pad.name: round(total / pad_full_scale, 5) for pad, total in per_pad
+            },
+            peak_taxel_counts=round(peak_rise(reading, tare.baseline), 1),
+            tactile_backend=tactile.name,
+            measured_at=timestamp(),
+        )
+
+
+def measure_tare(readings: list[TactileReading], spec: TactileSpec) -> Tare:
+    baseline = average_readings(readings)
+    noise_floor = rest_noise(readings, baseline, spec.full_scale_counts)
+
+    return Tare(
+        baseline=baseline,
+        noise_floor=noise_floor,
+        threshold=max(spec.contact_threshold, spec.noise_margin * noise_floor),
+    )
+
+
+def mean_counts(baseline: TactileBaseline) -> float:
+    total = sum(sum(pad) for pad in baseline.pads)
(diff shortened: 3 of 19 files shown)


mcp/gripper_mcp/datasheets/robotiq_2f_85.yaml
```diff
@@ -23,3 +23,24 @@ defaults:
...
+  # extension (hottest taxel ~106, no-contact floor ~0.0003 of full scale).
+  # Unverified against a physical TSF-85; re-measure before trusting it there.
+  full_scale_counts: 6200.0
+
+  # Contact is declared at or above this signal. Deliberately just above the
+  # noise floor: the point is to notice FIRST touch, not a firm grip.
+  contact_threshold: 0.02
+
+  # Matches the tactile SDK's findBaseline(), which averages 1000 samples.
+  baseline_samples: 1000
```
jlongvalRobotiq: the tsf is actually a product that is external and appended to the 2f grippers... im thinking it probably should have a file of its own.

mcp/tests/test_config.py
```diff
@@ -93,13 +94,44 @@ def test_a_default_effort_outside_the_rated_grip_force_is_rejected(tmp_path):
...
 
+def test_an_unknown_tactile_source_is_rejected(tmp_path):
+    path = wiring_file(
+        tmp_path,
+        [
+            {
+                "name": "left",
+                "model": "robotiq_2f_85",
+                "namespace": "/left",
+                "tactile": "sdk",
```
jlongvalRobotiq: this tactile source will eventually be present so im thinking we should change the counter example here.

mcp/gripper_mcp/datasheets/robotiq_2f_85.yaml
jlongvalRobotiq: Agreed, split out: `datasheets/tactile/robotiq_tsf_85.yaml` is the pads' own sheet, and `robotiq_2f_85.yaml` just names it with `tactile_model`. The loader stitches the two so `spec.tactile` is unchanged for consumers; a tactile block inlined in a gripper sheet is now rejected.

mcp/tests/test_config.py
jlongvalRobotiq: Changed to `tactile: mock`, the one source that will never be a wiring option (same counterexample as the gripper side).

mcp/gripper_mcp/tactile_service.py
```diff
@@ -0,0 +1,160 @@
...
+        return self._capture_baseline(gripper_name, tactile, spec)
+
+    def _capture_baseline(
+        self, gripper_name: str, tactile: TactileBackend, spec: TactileSpec
+    ) -> TactileBaseline:
+        baseline = average_readings(
+            [tactile.read_tactile() for _ in range(spec.baseline_samples)]
+        )
+        self._baselines[gripper_name] = baseline
+        return baseline
```
ebarnett3: ### 1. The tare averages 1000 samples that are all the same frame — High (with #66)

`_capture_baseline` (`tactile_service.py:114-121`) takes
`[tactile.read_tactile() for _ in range(spec.baseline_samples)]`, 1000 by
datasheet, and `average_readings` means them. That assumes each read is a fresh
sample. #66's source is explicit that it is not: "Reads return the latest
message received, never block on the topic rate"
(`ros_tactile_backend.py:5-7`), and `read_tactile` returns `self._static`
unchanged (`ros_tactile_backend.py:41-52`). A thousand Python-level reads
finish in a few milliseconds, far faster than frames arrive, so the mean is
mostly one frame repeated.

(That module's docstring says the sensor publishes at 60 Hz. Measured on the
bench it is about 2000 Hz, which changes the arithmetic below but not the
defect: the loop never waits for a new frame at either rate.)

```
caching (ros backend)    baseline rest per taxel = 100.0000
fresh sample per read    baseline rest per taxel = 102.9970
```

Confirmed on hardware: four tares took 21, 53, 131 and 604 ms, each reporting
`samples: 1000`. At 2000 Hz a thousand distinct frames take 0.5 s, so the 21 ms
call saw at most about 42 of them. The number in the response is not the number
averaged, and it changes call to call.

The tool still answers `samples: 1000` (`tactile_service.py:56`), so the report
says the noise was averaged out when it was not. `tactile.py:3-5` states the
intent plainly: "The baseline is the mean of many samples, matching the tactile
SDK's `findBaseline()`, which averages 1000."

Fix: collect distinct frames, and size the count to what the sensor can deliver.
At the measured 2000 Hz, 1000 real frames cost 0.5 s, which is fine for a tool
call, so the count itself can stay.

```python
def _capture_baseline(self, gripper_name, tactile, spec) -> TactileBaseline:
    baseline = average_readings(tactile.sample(spec.baseline_samples))
    self._baselines[gripper_name] = baseline
    return baseline
```

with `sample(count)` on the tactile protocol: the ROS source waits for a new
message between reads (a sequence counter or the message timestamp tells it
when one arrived) and gives up with a clear error if the topic stalls; the test
fake returns `count` scripted frames. `baseline_samples: 1000` then costs about
0.5 s of real averaging and the reported count is true, so the datasheet value
can stay as it is.

The alternative is #72: once the driver publishes baseline-subtracted rises,
the MCP stops baselining and this disappears. Worth deciding which before
shipping a tare tool that does not average.

mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
```diff
@@ -0,0 +1,25 @@
...
+# extension (hottest taxel ~106, no-contact floor ~0.0003 of full scale).
+# Unverified against a physical TSF-85; re-measure before trusting it there.
+full_scale_counts: 6200.0
+
+# Contact is declared at or above this signal. Deliberately just above the
+# noise floor: the point is to notice FIRST touch, not a firm grip.
+contact_threshold: 0.02
+
+# Matches the tactile SDK's findBaseline(), which averages 1000 samples.
+baseline_samples: 1000
```
ebarnett3, line 32: ### 2. On real pads the noise crosses the threshold, so `verify_grasp` answers `held` on an empty open gripper — Blocking

Measured on a real 2F-85 with TSF-85 pads, 2026-09-10, with the QoS fix from
#66 finding 1 applied so the tools receive data at all. Immediately after a
tare, gripper fully open, nothing touching either pad:

| read | `contact_signal` | `contact` (threshold 0.02) |
|---|---|---|
| 1 | 0.01685 | false |
| 2 | 0.03467 | **true** |
| 3 | 0.02891 | **true** |
| 4 | 0.02886 | **true** |

Then, still open and empty:

```
gripper_verify_grasp -> verdict='held', object_held=True, opening_mm=85.0,
                        contact_signal=0.03109, threshold=0.02,
                        detail='The pads register contact and the fingers stopped before meeting.'
```

The tool that exists to answer "am I holding something" says yes, with the
fingers wide open and nothing between them, within a second of being tared.

The constants come from the Isaac twin and the datasheet says so
(`datasheets/tactile/robotiq_tsf_85.yaml:14-22`: "Tuned against the Isaac Sim
twin's tactile extension ... Unverified against a physical TSF-85"). The real
scale is nothing like the simulated one: raw rest counts measure about 12540
per taxel against a simulated no-contact floor of 0.0003 of full scale, and
untouched `peak_taxel_counts` runs 17 to 24 counts, where `full_scale_counts:
6200.0` makes the 0.02 threshold equal about 124 counts summed over 56 taxels,
or roughly 2 counts each. Sensor noise is an order of magnitude above that.

Both numbers have to be measured on a physical sensor before these tools can
answer anything. Until then the honest options are to refuse the tactile tools
unless the datasheet is marked as measured, or to derive the threshold from the
tare's own observed spread rather than from a constant: the service already has
every rest sample at tare time, so the noise floor is free, and a threshold
below it can be rejected with the measured number in the message.

mcp/gripper_mcp/tactile_service.py
```diff
@@ -0,0 +1,160 @@
...
+    count = sum(len(pad) for pad in baseline.pads)
+    return total / count
+
+
+def judge_grasp(contact: bool, fingers_met: bool) -> GraspVerdict:
+    if fingers_met:
+        return "closed_on_nothing"
+    if contact:
+        return "held"
+    return "no_contact"
```
ebarnett3, line 201: ### 3. `judge_grasp` reaches the right internal-grip verdict for the wrong reason — Low (revised after the bench)

`judge_grasp` (`tactile_service.py:155-160`) reads a hold as contact on the pads
with the fingers apart. The TSF-85 pads sit on the fingers' gripping faces, the
ones that face each other. An internal grip holds a hollow object by pressing
outward with the *outside* of the fingers, which the manual documents as a
supported way to pick, so the pads face away from the object and register
nothing. The verdict is `no_contact` with `object_held: false` while the
gripper is holding the part.

Measured, that is not what happens, and I am correcting my own prediction. On a
real internal grip, fingers pressing outward on a bore at 49.05 mm:

```
gripper_read_tactile -> contact=True, contact_signal=0.10447, peak_taxel_counts=35.0
gripper_verify_grasp -> verdict='held', object_held=True
```

The pads do register something, about three times the empty-pad noise, whether
through finger deflection or contact at the bore edge. So the verdict is right.
It is right by accident: the fingers are apart and there is some signal, and the
detail text says "the fingers stopped before meeting" when they stopped while
opening. The margin is also thin, 0.104 against 0.035 of noise, where an
external hold gives 0.817, so it should not be relied on across objects.

Fix: say in the docstring and in `gripper_verify_grasp`'s description that an
internal hold is recognised weakly and the signal has a different origin, so
nobody tunes the threshold against it; and once gOBJ reaches the graph, gate on
it, since `DetectedWhileOpening` names an internal hold outright and needs no
inference from the pads at all.

##  `judge_grasp` on a closed gripper reports incorrect state

I judged this safe from the code and it is not. Measured on a real 2F-85 wearing
TSF-85 fingers, fully closed and empty, 1.5 s after the motion settled:

```
gripper_get_state   -> opening_mm=0.75, knuckle_rad=0.7859
gripper_read_tactile -> contact=True, contact_signal=0.2364
gripper_verify_grasp -> verdict='held', object_held=True,
                        detail='The pads register contact and the fingers stopped before meeting.'
```

`judge_grasp` (`tactile_service.py:155-160`) answers `closed_on_nothing` before
consulting the pads, which would save it, but only if `fingers_met` is true. The
gripper settles at 0.7859 rad, not the 0.7929 the description configures, most
likely because the TSF fingers are thicker than the standard ones and meet
earlier. That reads as 0.75 mm, `fingers_met`'s margin is 0.5 mm
(`service.py:179-180`), so the code believes the fingers stopped short, the pads
are pressing each other at 0.2364, and the verdict is `held` on an empty
gripper.

Two things follow. The margin has to cover the real closed opening with room to
spare, which 0.5 mm does not, and it should be measured per model rather than
shared with the motion classifier. And a threshold bump cannot substitute for
it: pads touching each other read 0.2364 against 0.817 for a real hold, so no
single contact threshold separates the two. Position has to make that call.

mcp/gripper_mcp/tactile_service.py
```diff
@@ -0,0 +1,160 @@
...
+
+        return TactileReadingResult(
+            gripper_name=gripper_name,
+            contact=signal >= spec.contact_threshold,
+            contact_signal=round(signal, 5),
+            threshold=spec.contact_threshold,
+            pad_signals={
+                pad.name: round(total / spec.full_scale_counts, 5)
+                for pad, total in per_pad
+            },
```
ebarnett3, line 172: ### 4. Per-pad signals are normalised against the two-pad full scale — Low

`tactile_service.py:139-142` divides each pad's own sum by
`spec.full_scale_counts`, which the datasheet defines as both pads at the
ceiling ("56 taxels x 110 counts", `datasheets/tactile/robotiq_tsf_85.yaml:14-18`).
A single pad pressed to saturation therefore reports 0.5, and the two `pad_signals`
sum to `contact_signal`. That is self-consistent but reads as a fraction of that
pad's own capacity, which it is not.

Fix: divide by `full_scale_counts / len(reading.pads)` and say in the model's
field description that each value is that pad's fraction of its own ceiling, or
keep the current scale and describe it as a share of the combined signal.

mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
```diff
@@ -0,0 +1,25 @@
...
+# The 7x4 grid per pad is what robotiq_tsf publishes (Taxels: uint16[28]).
+rows: 7
+cols: 4
+pads: [left, right]
+
+# Signal 1.0 means every taxel on both pads is at the working ceiling:
+# 56 taxels x 110 counts. Tuned against the Isaac Sim twin's tactile
+# extension (hottest taxel ~106, no-contact floor ~0.0003 of full scale).
+# Unverified against a physical TSF-85; re-measure before trusting it there.
+full_scale_counts: 6200.0
```
ebarnett3: ### 5. `full_scale_counts` does not match the arithmetic in its own comment — Nit

`datasheets/tactile/robotiq_tsf_85.yaml:14-18` explains the number as
"56 taxels x 110 counts", which is 6160, and then sets 6200.0. Use 6160.0, or
say the value is rounded and why.

mcp/gripper_mcp/tactile_service.py
```diff
@@ -0,0 +1,160 @@
...
+
+        return self._reading(gripper_name, tactile, baseline, spec)
+
+    def _pads_for(self, gripper_name: str) -> tuple[TactileBackend, TactileSpec]:
+        self._grippers.assert_known(gripper_name)
+        if gripper_name not in self._backends:
+            raise TactileUnavailableError(
+                f"Gripper '{gripper_name}' has no tactile pads configured."
+            )
+        return self._backends[gripper_name], self._specs[gripper_name]
```
ebarnett3: ### 6. `_pads_for` indexes `self._specs` unguarded — Nit

`tactile_service.py:96` returns `self._backends[gripper_name], self._specs[gripper_name]`
after checking only `_backends`. The two dicts are built together in #66's
`build_services`, so this is unreachable today; a `KeyError` here would be as
opaque as the one mbegin found in `GripperService._spec`. Index one dict of
`(backend, spec)` pairs and the question cannot arise.

mcp/gripper_mcp/tactile_service.py
```diff
@@ -0,0 +1,160 @@
...
+        specs: dict[str, TactileSpec],
+    ) -> None:
+        self._grippers = grippers
+        self._backends = backends
+        self._specs = specs
+        self._baselines: dict[str, TactileBaseline] = {}
+
+    def tare(self, gripper_name: str) -> TactileTareResult:
+        tactile, spec = self._pads_for(gripper_name)
+        baseline = self._capture_baseline(gripper_name, tactile, spec)
```
ebarnett3: ### 7. `_baselines` is mutated without a lock — Nit

`tactile_service.py:48, 120`. FastMCP runs tool calls concurrently, so a tare
and a read on the same gripper can interleave. The worst case is a wasted
capture rather than a wrong reading, since dict assignment is atomic, but a
`threading.Lock` around capture makes the intent explicit.

mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
```diff
@@ -0,0 +1,25 @@
+# Robotiq TSF-85 tactile fingers datasheet. The pads are a separate product
+# fitted onto a 2F-85 and read by their own driver (robotiq_tsf), so they get
+# their own sheet; the gripper's datasheet points here with `tactile_model`.
+# Whether a given gripper actually wears them is the wiring's business
```
ebarnett3: "Whether a given gripper actually wears them..."

This is a bit of a strange phrase. I would put

"Whether a given gripper has them or not..."

mcp/tests/test_config.py
```diff
@@ -93,13 +95,71 @@ def test_a_default_effort_outside_the_rated_grip_force_is_rejected(tmp_path):
         load_model_spec(write_datasheet(tmp_path, too_strong))
 
 
+def test_the_2f_85_datasheet_names_the_tsf_85_pads():
+    spec = load_model_spec(SPEC_DIR / "robotiq_2f_85.yaml")
+
+    assert spec.tactile_model == "robotiq_tsf_85"
+    assert spec.tactile.model == "robotiq_tsf_85"
```
ebarnett3: duplicated line

ebarnett3 (approved):

mcp/gripper_mcp/tactile_service.py
jlongvalRobotiq: Agreed, and the count was a lie on the ROS source. The tactile protocol now has `sample(count)`: count distinct frames, one per sensor update. The ROS source counts frames as they arrive and waits for the counter to move between reads, raising if the topic stalls for `STALE_FRAME_S`; the mock computes a fresh reading per call. `_capture_tare` uses it, so `samples: 1000` is now 1000 real frames, about 0.5 s at your measured 2 kHz. #72 stays the long-term answer; this keeps the tare honest until then.

mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
jlongvalRobotiq: Took your second option. The tare already reads the pads at rest, so the noise floor is measured every time: the runtime threshold is `max(contact_threshold, noise_margin x peak rest signal seen during the tare)`, with `noise_margin: 2.0` in the TSF-85 sheet. `gripper_tare_tactile` now reports `noise_floor` and the `threshold` in force, and every reading echoes that threshold. On your numbers (rest 0.035, internal hold 0.104, pads touching 0.236, external hold 0.817) it lands near 0.07, above the rest noise and below every contact case; a test replays those figures. The sheet keeps its "unverified" note. I have no TSF pads at the moment; #77 tracks the on-hardware check and the re-measured constants, assigned to me. Resolving here.

mcp/gripper_mcp/tactile_service.py
jlongvalRobotiq: Both halves taken. Docstring and `gripper_verify_grasp`'s description now say an internal hold is recognised only weakly and from a different origin, so nobody tunes against it; gating on gOBJ (`DetectedWhileOpening`) is the planned replacement once the backend reads it. The closed case is fixed on the position side in #61 (merged): `closed_tolerance_mm` is 1.5 mm per model, above two position counts and covering the 0.75 mm short settle you measured with TSF fingers. `judge_grasp` now answers `closed_on_nothing` there.

mcp/gripper_mcp/tactile_service.py
jlongvalRobotiq: Each pad is now divided by `full_scale_counts / len(pads)`, so one saturated pad reads 1.0 on its own line while `contact_signal` stays normalised across both. Field description and module docstring say so.

mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
jlongvalRobotiq: 6160.0.

mcp/gripper_mcp/tactile_service.py
jlongvalRobotiq: One dict of `(backend, spec)` pairs now, built as such in `build_services`.

mcp/gripper_mcp/tactile_service.py
jlongvalRobotiq: Capture runs under a `threading.Lock`.

mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
jlongvalRobotiq: "has them or not".

mcp/tests/test_config.py
jlongvalRobotiq: Dropped.

PR #66, opened 2026-09-04
feat(mcp): ros tactile source on robotiq_tsf

jlongvalRobotiq:
## Change

Tenth PR of the `mcp/` series (stacked on #65). Tactile reaches the driver:

1. **`ros_tactile_backend.py`** — `RosTactileBackend` subscribes to `TactileSensor/StaticData` with the sensor-data QoS (the driver publishes best effort; a reliable subscriber never matches it and simply hears nothing), under the gripper's namespace or the wiring's `tactile_namespace` when the `robotiq_tsf` driver runs elsewhere, on the executor thread the gripper backends already share. A read returns the latest frame and never blocks on the topic rate (about 2 kHz on the bench), because a contact loop reads far more often than it moves; it refuses a frame older than `STALE_FRAME_S` so a driver that stopped publishing cannot keep answering. `sample(count)` is the exception: the callback buffers `count` consecutive frames and hands them over once, so a tare averages distinct frames and reports a true count (see the live block for what that costs). The `Dynamic` channel is not read: nothing in the series consumes it.
2. **`ros_tactile_messages.py`** — shapes `StaticData` into a `TactileReading` with no ROS imports, so the pad-count and grid mismatches it reports, and the stale-frame and stalled-sample messages, are unit-tested in CI. Counts stay raw; `tactile.py` remains the only place baseline arithmetic happens.
3. **`tactile: ros`** in `grippers.yaml` is now built by the default factory; the import is lazy so the unit tests run without ROS. The example wiring puts TSF-85 pads on the left 2F-85.

The tactile-guided closing loop (`gripper_grasp_until_contact`) is split out to its own PR, stacked on this one, until it has run on a real gripper with pads.

## Verification

- `uv run pytest`: 189 passed (7 new): message shaping (two named pads, one-pad / short-grid mismatches, stale-frame and stalled-sample messages); `tactile: ros` without rclpy names `robotiq_tsf`; `tactile_namespace` defaults to the gripper's namespace; the example wiring loads with one tactile entry.
- **Live, in `robotiq_ros2:jazzy`** (rerun after this round): driver on fake hardware + a dummy `robotiq_tsf` publisher (real `robotiq_tsf/msg/StaticData`, sensor-data QoS, ~1.7 kHz, rest 12540 ± 2 counts per taxel so the tare has real noise to average, pressure following the fake gripper's `joint_states` with a 46 mm virtual object), `RosGripperBackend` + `RosTactileBackend` through the real services. "Close on the object" is `move_to_opening(45.0)` because fake hardware cannot stall on the virtual object; a full `close_fully` reaches 0.0 and reads `closed_on_nothing` with the pads touching at ~0.999.

  ```
  list:         [{'gripper_name': 'left', 'model': 'robotiq_2f_85', 'backend': 'ros', 'max_opening_mm': 85.0, 'tactile': 'ros'}]
  open:         reached
  tare:         {'samples': 1000, 'rest_counts_mean': 12540.0, 'noise_floor': 0.00866, 'threshold': 0.02, 'tactile_backend': 'ros'} wall=8.81s
  read open:    {'contact': False, 'contact_signal': 0.00407, 'threshold': 0.02, 'pad_signals': {'left': 0.00363, 'right': 0.00452}, 'peak_taxel_counts': 2.1}
  close 45mm:   reached
  verify:       {'verdict': 'held', 'object_held': True, 'opening_mm': 45.0, 'contact_signal': 0.26449, 'threshold': 0.02}
  open:         reached
  verify open:  no_contact
  --- ros2 topic info -v /TactileSensor/StaticData
  Publisher    dummy_tsf                 Reliability: BEST_EFFORT
  Subscription gripper_mcp_tactile_left  Reliability: BEST_EFFORT
  --- stale check: dummy publisher stopped, 3 s later
  read:         RuntimeError: The last TactileSensor/StaticData frame is 3.0 s old (limit 2 s); the robotiq_tsf driver has stopped publishing.
  ```

  The tare's 1000 distinct frames took 8.8 s, not the 0.5 s the publish rate allows: rclpy's executor delivers about 110 `StaticData` callbacks per second here whatever the publisher does (measured the same with a per-frame wake-up and with the callback buffering the whole sample). Reported truthfully in `samples`; whether 1000 is the right count for a tool call is open.

## Stack
Stacked on #65: this PR shows only its own 2 commits.

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -49,9 +49,11 @@ hold. `gripper_verify_grasp` gives one of:
 | `closed_on_nothing` | Fingers fully closed |
 | `no_contact` | Fingers apart, pads quiet: the object slipped, or the stop was outside the pads |
 
-The ROS source on `robotiq_tsf`'s `TactileSensor/StaticData` topic is next;
-until it lands, `build_services` refuses a `tactile` entry by name. The tests
-drive the tools from a scripted pad model under `tests/fakes/`.
+The tactile source subscribes to `robotiq_tsf`'s `TactileSensor/StaticData`
+under the gripper's `namespace` (the tactile driver runs in the same namespace
+as the gripper's controller); it needs `robotiq_tsf` built
+and sourced alongside `rclpy`. The tests drive the tools from a scripted pad
+model under `tests/fakes/`.
 
 ## Quick start
 
--- a/mcp/gripper_mcp/ros_tactile_backend.py
+++ b/mcp/gripper_mcp/ros_tactile_backend.py
@@ -1,0 +1,82 @@
+"""Tactile readings from a TSF-85 through this repo's robotiq_tsf driver.
+
+One node per gripper, in the namespace the wiring gives the pads (the gripper's
+own unless `tactile_namespace` says otherwise), so `TactileSensor/StaticData`
+and `TactileSensor/Dynamic` resolve to the pads on that gripper. It shares the
+executor thread with the gripper backends (RosGraph).
+
+The driver publishes with the sensor-data QoS (best effort), so the
+subscriptions must too: a reliable subscriber never matches a best-effort
+publisher and simply receives nothing, with no error anywhere.
+
+Reads return the latest message received and never block on the topic rate:
+the sensor publishes at about 2 kHz, and a contact loop reads far more often
+than it moves. Because of that, a read remembers when the last frame arrived
+and refuses to answer from a frame older than `STALE_FRAME_S`; otherwise a
+driver that stopped publishing would keep reporting its last frame forever,
+and a frozen "no contact" reads as "keep closing" to a tactile-guided close.
+"""
+
+import time
+
+from rclpy.node import Node
+from rclpy.qos import qos_profile_sensor_data
+from robotiq_tsf.msg import Dynamic, StaticData
+
+from gripper_mcp.ros_backend import RosGraph, poll_until
+from gripper_mcp.ros_tactile_messages import (
+    DYNAMIC_TOPIC,
+    STATIC_TOPIC,
+    reading_from_counts,
+    stale_frame_message,
+)
+from gripper_mcp.tactile_backend import TactileLayout, TactileReading
+
+STATIC_WAIT_S = 2.0
+STALE_FRAME_S = 2.0
+
+
+class RosTactileBackend:
+    name = "ros"
+
+    def __init__(
+        self, gripper_name: str, namespace: str, layout: TactileLayout
+    ) -> None:
+        self._layout = layout
+        executor = RosGraph.executor()
+        self._node = Node(
+            f"gripper_mcp_tactile_{gripper_name}", namespace=namespace or "/"
+        )
+        self._static: StaticData | None = None
+        self._dynamic: Dynamic | None = None
+        self._latest_at = 0.0
+        self._node.create_subscription(
+            StaticData, STATIC_TOPIC, self._on_static, qos_profile_sensor_data
+        )
+        self._node.create_subscription(
+            Dynamic, DYNAMIC_TOPIC, self._on_dynamic, qos_profile_sensor_data
+        )
+        executor.add_node(self._node)
+
+    def read_tactile(self) -> TactileReading:
+        static = poll_until(lambda: self._static, STATIC_WAIT_S)
+        dynamic = self._dynamic
+        if static is None:
+            raise RuntimeError(
+                f"No {STATIC_TOPIC} under '{self._node.get_namespace()}' within "
+                f"{STATIC_WAIT_S:.0f} s; is the robotiq_tsf driver running?"
+            )
+        age_s = time.monotonic() - self._latest_at
+        if age_s > STALE_FRAME_S:
+            raise RuntimeError(stale_frame_message(age_s, STALE_FRAME_S))
+        dynamics = [] if dynamic is None else [d.value for d in dynamic.data]
+        return reading_from_counts(
+            [list(pad.values) for pad in static.taxels], dynamics, self._layout
+        )
+
+    def _on_static(self, message: StaticData) -> None:
+        self._static = message
+        self._latest_at = time.monotonic()
+
+    def _on_dynamic(self, message: Dynamic) -> None:
+        self._dynamic = message
--- a/mcp/gripper_mcp/ros_tactile_backend.py
+++ b/mcp/gripper_mcp/ros_tactile_backend.py
@@ -2,11 +2,11 @@
 
 One node per gripper, in the namespace the wiring gives the pads (the gripper's
 own unless `tactile_namespace` says otherwise), so `TactileSensor/StaticData`
-and `TactileSensor/Dynamic` resolve to the pads on that gripper. It shares the
-executor thread with the gripper backends (RosGraph).
+resolves to the pads on that gripper. It shares the executor thread with the
+gripper backends (RosGraph).
 
 The driver publishes with the sensor-data QoS (best effort), so the
-subscriptions must too: a reliable subscriber never matches a best-effort
+subscription must too: a reliable subscriber never matches a best-effort
 publisher and simply receives nothing, with no error anywhere.
 
 Reads return the latest message received and never block on the topic rate:
@@ -15,20 +15,28 @@ than it moves. Because of that, a read remembers when the last frame arrived
 and refuses to answer from a frame older than `STALE_FRAME_S`; otherwise a
 driver that stopped publishing would keep reporting its last frame forever,
 and a frozen "no contact" reads as "keep closing" to a tactile-guided close.
+
+`sample` is the exception: it asks the subscription callback to keep the next
+`count` messages and sleeps until the callback signals the buffer is full, so
+each frame is a distinct sensor update and the two threads hand off once per
+sample rather than once per frame (a per-frame handoff under the GIL was
+measured at under 100 frames/s against a 2 kHz publisher). It gives up with a
+stalled-sample error if no frame arrives for `STALE_FRAME_S` mid-sample.
 """
 
+import threading
 import time
 
 from rclpy.node import Node
 from rclpy.qos import qos_profile_sensor_data
-from robotiq_tsf.msg import Dynamic, StaticData
+from robotiq_tsf.msg import StaticData
 
 from gripper_mcp.ros_backend import RosGraph, poll_until
 from gripper_mcp.ros_tactile_messages import (
-    DYNAMIC_TOPIC,
     STATIC_TOPIC,
     reading_from_counts,
     stale_frame_message,
+    stalled_sample_message,
 )
 from gripper_mcp.tactile_backend import TactileLayout, TactileReading
 
@@ -48,19 +56,17 @@ class RosTactileBackend:
             f"gripper_mcp_tactile_{gripper_name}", namespace=namespace or "/"
         )
         self._static: StaticData | None = None
-        self._dynamic: Dynamic | None = None
         self._latest_at = 0.0
+        self._wanted = 0
+        self._collected: list[StaticData] = []
+        self._sample_done = threading.Event()
         self._node.create_subscription(
             StaticData, STATIC_TOPIC, self._on_static, qos_profile_sensor_data
-        )
-        self._node.create_subscription(
-            Dynamic, DYNAMIC_TOPIC, self._on_dynamic, qos_profile_sensor_data
         )
         executor.add_node(self._node)
 
     def read_tactile(self) -> TactileReading:
         static = poll_until(lambda: self._static, STATIC_WAIT_S)
-        dynamic = self._dynamic
         if static is None:
             raise RuntimeError(
                 f"No {STATIC_TOPIC} under '{self._node.get_namespace()}' within "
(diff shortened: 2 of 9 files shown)


mcp/README.md
```diff
@@ -49,9 +50,17 @@ automatically if the gripper is open; `gripper_tare_tactile` retakes it).
...
 | `no_contact` | Fingers apart, pads quiet: the object slipped, or the stop was outside the pads |
 
-The ROS source on `robotiq_tsf`'s `TactileSensor/StaticData` topic is next;
-until it lands, `build_services` refuses a `tactile` entry by name. The tests
-drive the tools from a scripted pad model under `tests/fakes/`.
+`gripper_grasp_until_contact` is the one tool that moves more than once per
+call: it closes by the datasheet's `step_mm`, reads the pads between steps, and
+stops on first touch, on the fingers meeting, on a stall, or on a hard
+`contact_timeout_s`. A stall with quiet pads is reported as
+`stalled_before_contact`, never as a grasp.
```
jlongvalRobotiq: i havent tested this functionality on a real gripper so im not sure we can add it at the moment . we should make this PR be only about robotiq_ros_tsf.

jlongvalRobotiq: Done: this PR is now the robotiq_tsf source and its wiring only (156 tests). `gripper_grasp_until_contact` moved unchanged to #73, stacked on this one, and stays there until it has run on a real gripper with pads.

mcp/gripper_mcp/ros_tactile_backend.py
```diff
@@ -0,0 +1,64 @@
...
+        self._node = Node(
+            f"gripper_mcp_tactile_{gripper_name}", namespace=namespace or "/"
+        )
+        self._static: StaticData | None = None
+        self._dynamic: Dynamic | None = None
+        self._node.create_subscription(StaticData, STATIC_TOPIC, self._on_static, 10)
+        self._node.create_subscription(Dynamic, DYNAMIC_TOPIC, self._on_dynamic, 10)
+        executor.add_node(self._node)
+
+    def read_tactile(self) -> TactileReading:
```
mbegin-robotiq, line 68: A read returning the latest cached message is the right call for a contact loop, but it silently guts the tare.

`TactileService._capture_baseline` does `[tactile.read_tactile() for _ in range(spec.baseline_samples)]` with `baseline_samples: 1000`. Against a cached source that loop never waits, so it averages the same frame a thousand times. Measured against a 60 Hz publisher, driving this class's real `_on_static` / `read_tactile` path:

```
1000 read_tactile() calls took 11.1 ms
driver published ~1 frames in that window
distinct pad frames actually averaged: 4  (of 1000 samples requested)
time needed for 1000 genuinely distinct samples at 60 Hz: 16.7 s
```

The baseline exists to suppress noise — `tactile.py` says so, and the datasheet comment says it "Matches the tactile SDK's findBaseline(), which averages 1000 samples". Averaging one frame a thousand times suppresses nothing, and `TactileTareResult.samples` then reports 1000 to the agent. The live output in the description shows the symptom: `rest_counts_mean=9.0`, the round number a single unaveraged frame gives.

This is specific to this PR: the mock source from #62 returns a fresh reading per call, so the tare behaved as designed until the cached ROS source replaced it. Nothing tests the combination.

Needs a decision rather than a one-liner — a blocking "next frame" read would make a 1000-sample tare take 16.7 s, so probably fewer samples over a bounded window (say 60 over 1 s), with `samples` reporting frames actually averaged.

mcp/gripper_mcp/ros_tactile_backend.py
```diff
@@ -0,0 +1,64 @@
...
+            [list(pad.values) for pad in self._static.taxels], dynamics, self._layout
+        )
+
+    def _on_static(self, message: StaticData) -> None:
+        self._static = message
+
+    def _on_dynamic(self, message: Dynamic) -> None:
+        self._dynamic = message
+
+    def _await_static(self) -> bool:
```
mbegin-robotiq: Two things here.

This hand-rolls the loop that `poll_until` already implements, in the same module this line's imports (`POLL_S`, `STATE_WAIT_S`) come from — `poll_until(lambda: self._static, STATE_WAIT_S)` is the same thing.

More importantly, nothing records *when* the message arrived, so `read_tactile` serves the last frame at any age. `RosGripperBackend` tracks `_latest_at` and reports `STALE_STATE_S` in `health()`; there is no equivalent here. If `robotiq_tsf` stops publishing mid-session this backend keeps answering with the last frame forever, indistinguishable from a live sensor.

That matters most for #73's tactile-guided closing loop: a frozen `contact=False` reads as "keep closing".

mcp/gripper_mcp/ros_tactile_backend.py
```diff
@@ -0,0 +1,64 @@
...
+        self._node.create_subscription(Dynamic, DYNAMIC_TOPIC, self._on_dynamic, 10)
+        executor.add_node(self._node)
+
+    def read_tactile(self) -> TactileReading:
+        if not self._await_static():
+            raise RuntimeError(
+                f"No {STATIC_TOPIC} under '{self._node.get_namespace()}' within "
+                f"{STATE_WAIT_S:.0f} s; is the robotiq_tsf driver running?"
+            )
+        dynamics = (
```
mbegin-robotiq: Minor: `self._static` is read here and again on the `reading_from_counts` line below, with `self._dynamic` in between, while the executor thread can replace either. Worth snapshotting both into locals at the top of the read so one call cannot mix taxels from one frame with a dynamic value from another.

Unlikely to bite given the 60 Hz rate, but it is free to make it impossible.

mbegin-robotiq (approved):

mcp/gripper_mcp/ros_tactile_backend.py
```diff
@@ -0,0 +1,64 @@
...
+        self, gripper_name: str, namespace: str, layout: TactileLayout
+    ) -> None:
+        self._layout = layout
+        executor = RosGraph.executor()
+        self._node = Node(
+            f"gripper_mcp_tactile_{gripper_name}", namespace=namespace or "/"
+        )
+        self._static: StaticData | None = None
+        self._dynamic: Dynamic | None = None
+        self._node.create_subscription(StaticData, STATIC_TOPIC, self._on_static, 10)
```
ebarnett3: ### 1. The subscriptions are RELIABLE against a BEST_EFFORT publisher, so no tactile data ever arrives — Blocking

Measured on a real 2F-85 with TSF-85 pads, 2026-09-10. Every tactile tool fails:

```
gripper_tare_tactile -> No TactileSensor/StaticData under '/' within 2 s; is the robotiq_tsf driver running?
gripper_read_tactile -> (same)
gripper_verify_grasp -> (same)
```

The driver was running and publishing the whole time. The cause is a QoS
incompatibility. Both subscriptions pass a plain depth
(`ros_tactile_backend.py:37` for `StaticData`, `:38` for `Dynamic`), which is
RELIABLE by default:

```python
37    self._node.create_subscription(StaticData, STATIC_TOPIC, self._on_static, 10)
38    self._node.create_subscription(Dynamic, DYNAMIC_TOPIC, self._on_dynamic, 10)
```

The driver publishes both with `rclcpp::SensorDataQoS()`, which is BEST_EFFORT
(`robotiq_tsf/src/poll_data_sdk.cpp:136-138`):

```cpp
136    auto qos = rclcpp::SensorDataQoS();
137    static_pub_ = create_publisher<msg::StaticData>("TactileSensor/StaticData", qos);
138    dynamic_pub_ = create_publisher<msg::Dynamic>("TactileSensor/Dynamic", qos);
```

A RELIABLE subscriber never matches a BEST_EFFORT publisher, so the callbacks
are never invoked, `self._static` stays `None`, and `_await_static`
(`ros_tactile_backend.py:60-64`) times out on every call.

```
$ ros2 topic info /TactileSensor/StaticData --verbose
Publisher count: 1     Node name: poll_data                 Reliability: BEST_EFFORT
Subscription count: 1  Node name: gripper_mcp_tactile_left  Reliability: RELIABLE
```

Fix, verified on the bench:

```python
from rclpy.qos import qos_profile_sensor_data
...
self._node.create_subscription(
    StaticData, STATIC_TOPIC, self._on_static, qos_profile_sensor_data
)
self._node.create_subscription(
    Dynamic, DYNAMIC_TOPIC, self._on_dynamic, qos_profile_sensor_data
)
```

With that change the tools return data immediately. Nothing in the suite could
have caught this: the MCP tests never stand up a ROS graph, and
`test_ros_tactile_messages.py` exercises the pure adapter. A single end-to-end
check against the driver is what this needs, which is the same gap #61's
finding 4 describes for the description.

mcp/gripper_mcp/server.py
```diff
@@ -107,10 +107,18 @@ def build_tactile(
             f"Gripper '{config.name}' asks for tactile pads, but model "
             f"'{config.model}' names no tactile_model in its datasheet."
         )
-    raise NotImplementedError(
-        f"Gripper '{config.name}' asks for the ROS tactile source, which is not in "
-        "this build."
-    )
+    return build_ros_tactile(config, spec)
```
ebarnett3, line 113: ### 3. `build_tactile` takes a gripper backend it never uses — Nit

`server.py:100-102` accepts `gripper: GripperBackend` and neither branch touches
it (`server.py:103-110`). If it is there so a future source can share the
gripper's node, say so; otherwise drop it from the signature and from
`TactileFactory`.

mcp/gripper_mcp/ros_tactile_backend.py
```diff
@@ -0,0 +1,64 @@
...
+message received, never block on the topic rate: the sensor publishes at 60 Hz,
+and a contact loop reads far more often than it moves.
+"""
+
+import time
+
+from rclpy.node import Node
+from robotiq_tsf.msg import Dynamic, StaticData
+
+from gripper_mcp.ros_backend import POLL_S, STATE_WAIT_S, RosGraph
```
ebarnett3: ### 4. The tactile source imports its timing constants from the gripper backend — Nit

`ros_tactile_backend.py:15` pulls `POLL_S` and `STATE_WAIT_S` out of
`ros_backend`, so the pads' wait budget is defined by the gripper's module and
changes with it. They are independent devices with different publish rates, and
the bench measured 2000 Hz on the pads against the controller's much lower
rate. Give this module its own `STATIC_WAIT_S` and `POLL_S`. Partly overlaps
mbegin's `:60` thread, which notes the same imports while making a different
point.

mcp/gripper_mcp/server.py
```diff
@@ -107,10 +107,18 @@ def build_tactile(
...
+
+def build_ros_tactile(config: GripperConfig, spec: GripperModelSpec) -> TactileBackend:
+    try:
+        from gripper_mcp.ros_tactile_backend import RosTactileBackend
+    except ImportError as error:
+        raise RuntimeError(
+            f"Gripper '{config.name}' uses the ros tactile source, which needs a "
+            f"sourced ROS 2 install with rclpy and robotiq_tsf: {error}"
+        ) from error
+    return RosTactileBackend(config.name, config.namespace, spec.tactile.layout)
```
ebarnett3: ### 2. The pads must share the gripper's namespace, with nothing to override it — Medium

`build_ros_tactile` passes `config.namespace` straight through
(`server.py:121`), and `GripperConfig` has only the one namespace field
(`config.py:48, 50`). So the TSF driver has to run in the same namespace as
`robotiq_gripper_controller`. Nothing in this repo makes that true: `tsf_driver.launch.py`
brings the driver up on its own, and a cell that namespaces the pads by sensor
(`/left_tsf`) or leaves them at the root while the gripper is under `/left`
cannot be wired at all, with the failure showing up as "No
TactileSensor/StaticData under '/left'".

Fix: let the wiring say where the pads are, defaulting to the gripper's
namespace so the common case stays a single word.

```python
class GripperConfig(StrictModel):
    name: str
    model: str
    namespace: str
    description: str = ""
    tactile: Literal["ros"] | None = None
    tactile_namespace: str | None = None  # defaults to `namespace`
```

with `build_ros_tactile` using `config.tactile_namespace or config.namespace`,
and a line in `grippers.yaml.example` next to the `tactile: ros` comment.

ebarnett3 (approved):

mcp/gripper_mcp/ros_tactile_backend.py
jlongvalRobotiq: Decided on #65 with Eric's matching thread: the tactile protocol gained `sample(count)`, and this source implements it by having the callback buffer `count` consecutive frames, so every frame in the tare is a distinct sensor update, and it raises with "after N of M samples" if the driver stops mid-tare. Eric measured the pads at about 2 kHz, not 60 Hz. The catch, from the rerun in the description: rclpy delivers only ~110 `StaticData` callbacks/s to Python whatever the publisher rate, so 1000 distinct frames take ~9 s, not the 0.5 s the rate suggests. The count is now true and the noise is really averaged; whether 1000 is the right number for a ~9 s tool call, or your 60-over-1 s is, is the open question. I lean to fewer samples in a follow-up rather than blocking this PR on it. `read_tactile` keeps returning the latest frame for the contact loop.

mcp/gripper_mcp/ros_tactile_backend.py
jlongvalRobotiq: `poll_until` now, with this module's own `STATIC_WAIT_S`. Arrival time is tracked in `_latest_at` and a read refuses a frame older than `STALE_FRAME_S` (2 s) with "the robotiq_tsf driver has stopped publishing", so a frozen `contact=False` cannot feed #73's loop. The stale path is in the live block.

mcp/gripper_mcp/ros_tactile_backend.py
jlongvalRobotiq: The `Dynamic` subscription is gone (dropped on #62, no consumer), so a read now takes one `static` snapshot and derives everything from it.

mcp/gripper_mcp/ros_tactile_backend.py
jlongvalRobotiq: `qos_profile_sensor_data` on the subscription, and the module docstring says why, since nothing errors when the QoS does not match. Thanks for the bench confirmation; the live block now shows both ends BEST_EFFORT. The end-to-end gap against the real driver stays on the list with #61's.

mcp/gripper_mcp/server.py
jlongvalRobotiq: `GripperConfig.tactile_namespace`, defaulting to the gripper's namespace, used by `build_ros_tactile`; documented in `grippers.yaml.example` next to `tactile: ros`, tested in `test_config`.

mcp/gripper_mcp/server.py
jlongvalRobotiq: Dropped from `build_tactile` and `TactileFactory`.

mcp/gripper_mcp/ros_tactile_backend.py
jlongvalRobotiq: Own `STATIC_WAIT_S` and `STALE_FRAME_S` here; nothing is imported from `ros_backend` but `RosGraph` and `poll_until`.

mcp/gripper_mcp/ros_tactile_backend.py
ebarnett3: I agree that 9 s is too long. 1 s seems like a good compromise.

PR #69, opened 2026-09-04
build(mcp): Docker image and compose service for the gripper MCP server

jlongvalRobotiq:
Step 10 of the `mcp/` ladder: the server as a container. #66 merged, so this now sits directly on `feat/mcp`.

## What

1. **`mcp/Dockerfile`** — `FROM` the workspace image that `docker/Dockerfile` already builds (`BASE_IMAGE` arg, default `robotiq_ros2:jazzy`). That image has `rclpy`, `control_msgs` and `robotiq_tsf`'s messages on the distro's Python; uv installs `mcp/` into a venv on that same interpreter with system site-packages visible, so one Python imports both `rclpy` and `fastmcp`. `uv sync --locked --no-dev` from `uv.lock`. Both `rmw_fastrtps_cpp` and `rmw_cyclonedds_cpp` installed so `RMW_IMPLEMENTATION` picks the graph at run time.
2. **`mcp/entrypoint.sh`** — sources ROS and the workspace, execs `gripper-mcp` with the container's arguments. `grippers.yaml` is mounted at `/config/grippers.yaml`, never baked in.
3. **`docker/docker-compose.yml`** — one `gripper-mcp` service on the host network (shared DDS graph), with `ROBOTIQ_ROS2_IMAGE`, `GRIPPER_MCP_IMAGE` (the tag the build lands on, so a lyrical build cannot overwrite the jazzy tag), `GRIPPERS_YAML`, `ROS_DOMAIN_ID`, `RMW_IMPLEMENTATION`, `GRIPPER_MCP_PORT` as optional knobs. `GRIPPERS_YAML` defaults to the shipped example so a fresh checkout starts instead of mounting a root-owned directory; the header says it is a compose bind source (absolute or `./`-prefixed). The base image is built first; compose cannot order one service's build after another's.
4. **READMEs** — Docker section in `mcp/README.md`; pointer from the root README's Docker section.

## Why a separate Dockerfile, not a third stage

A third stage in `docker/Dockerfile` would become the default build target and change what `run.sh` and the documented `docker build` produce. A `BASE_IMAGE` arg keeps `docker/Dockerfile` untouched and lets a headless or downstream build of it be the base.

## Verified

- Image builds from `robotiq_ros2:jazzy` (1.68 GB vs 1.37 GB base); `/opt/gripper_mcp/bin/python` imports `rclpy`, `robotiq_tsf.msg`, `fastmcp` and `gripper_mcp`.
- ROS wiring: driver on `use_fake_hardware` in one container, this image in another, both `--network host` on domain 77. 11 tools listed; `gripper_get_health` reports the `ParallelGripperCommand` action ready; `gripper_set_opening` 30 mm and `gripper_open` both `reached`.
- `docker compose -f docker/docker-compose.yml up` with `GRIPPERS_YAML`/`GRIPPER_MCP_PORT` overrides: service starts, tool call answers, `down` cleans up.
- pre-commit clean; hadolint clean apart from apt version pinning (same as `docker/Dockerfile`).
- From a fresh shallow clone of this branch (no `mcp/grippers.yaml`), 2026-09-16: base image built from `docker/Dockerfile` (3 min), `docker compose -f docker/docker-compose.yml up --build` on defaults mounts the example wiring and answers on `:8300/mcp` in ~6 s; `demo.sh` on the same images brings the three containers up and answers in ~6 s; both `down` clean.
- Demo rerun after the review round (2026-09-16, image rebuilt from this branch): `demo.sh up` with a missing tag exits 1 and prints the two build commands; with both images, all three containers up, `initialize` on `:8301/mcp` answers `robotiq_gripper_mcp` after ~6 s; with the host exporting `ROS_DOMAIN_ID=100`, all three containers report 42; `demo.sh down` removes the containers and the `xhost` grant.

## Not in this PR

No CI job builds the image: it needs the multi-minute workspace image first. Can be added to `ci-ros-build-test.yml` if wanted.

## Demo

Everything running with no hardware (`mcp/demo/`): the driver on ros2_control's fake hardware with RViz showing the 2F-85, and the server with one gripper wired, `driver`, at the root of the graph (`namespace: /`). `ROS_DOMAIN_ID` is fixed at 42 in the demo compose, not read from the environment, so the demo never joins a real cell's graph. `demo.sh` checks both images exist before `up` and grants X only to `local:root`, revoked on `down`.

```bash
./mcp/demo/demo.sh
claude mcp add --transport http robotiq-gripper http://127.0.0.1:8301/mcp
```

[Screencast from 2026-09-04 13-48-56.webm](https://github.com/user-attachments/assets/697cb565-3a31-459a-8bf5-ed846a076ace)


mcp/demo/docker-compose.yml
```diff
@@ -0,0 +1,68 @@
...
+# ROS_DOMAIN_ID defaults to 42 and the port to 8301 so the demo never joins a
+# real cell's graph or shadows a server already on 8300.
+services:
+  gripper-driver:
+    image: ${ROBOTIQ_ROS2_IMAGE:-robotiq_ros2:jazzy}
+    container_name: robotiq_demo_driver
+    network_mode: host
+    ipc: host
+    environment:
+      ROS_DOMAIN_ID: ${ROS_DOMAIN_ID:-42}
```
mbegin-robotiq: **The demo can join a real cell's graph.**

The header (l.13-14) promises "ROS_DOMAIN_ID defaults to 42 … so the demo never joins a real cell's graph", but `${ROS_DOMAIN_ID:-42}` falls back to 42 only when the variable is *unset*. On a real cell `ROS_DOMAIN_ID` is normally exported from the shell profile, so all three demo containers silently join that domain instead.

The fake-hardware driver then publishes onto the live graph, and the demo MCP server can bind to the *real* driver's action server — `gripper_open` on the demo moves real hardware. Two `controller_manager`/`robot_state_publisher` pairs on one domain is its own mess.

Hard-coding `42`, or reading a separate `DEMO_ROS_DOMAIN_ID`, makes the guarantee in the header true.

docker/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    image: robotiq_gripper_mcp:jazzy
+    container_name: robotiq_gripper_mcp
+    network_mode: host
+    ipc: host
+    restart: unless-stopped
+    environment:
+      ROS_DOMAIN_ID: ${ROS_DOMAIN_ID:-0}
+      RMW_IMPLEMENTATION: ${RMW_IMPLEMENTATION:-rmw_fastrtps_cpp}
+    volumes:
+      - ${GRIPPERS_YAML:-../mcp/grippers.yaml}:/config/grippers.yaml:ro
```
mbegin-robotiq: **Default bind source doesn't exist in a fresh checkout.**

`../mcp/grippers.yaml` is gitignored, so no clone has it. The header does say to `cp` first, but if that step is skipped Docker creates a root-owned *directory* at `mcp/grippers.yaml`. The server then crash-loops under `restart: unless-stopped`, and the documented `cp mcp/grippers.yaml.example mcp/grippers.yaml` recovery copies *into* that directory rather than failing cleanly — so the obvious fix doesn't fix it.

Defaulting to `../mcp/grippers.yaml.example` would start with the example wiring and degrade far more gracefully.

docker/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+# Then point an MCP client at http://127.0.0.1:8300/mcp. Environment knobs,
+# all optional: ROBOTIQ_ROS2_IMAGE (base image), GRIPPERS_YAML (wiring file),
+# ROS_DOMAIN_ID, RMW_IMPLEMENTATION, GRIPPER_MCP_PORT.
+services:
+  gripper-mcp:
+    build:
+      context: ../mcp
+      args:
+        BASE_IMAGE: ${ROBOTIQ_ROS2_IMAGE:-robotiq_ros2:jazzy}
+    image: robotiq_gripper_mcp:jazzy
```
mbegin-robotiq: **Tag is hard-coded jazzy while the base is a variable.**

`BASE_IMAGE` may be `robotiq_ros2:lyrical`, but the built image always lands on `robotiq_gripper_mcp:jazzy`. A lyrical build therefore overwrites the jazzy tag that both the README's `docker run` and `mcp/demo/docker-compose.yml`'s default (`${GRIPPER_MCP_IMAGE:-robotiq_gripper_mcp:jazzy}`) point at — the demo would then silently run the lyrical image.

Deriving the tag from the same variable, or at least a `GRIPPER_MCP_IMAGE` knob mirroring the demo compose, would keep the two in step.

docker/docker-compose.yml
mbegin-robotiq: **`GRIPPERS_YAML` carries compose bind-source semantics, not path semantics.**

Two sharp edges for something advertised as a plain knob in the header:

- a value without a leading `./` or `/` (say `GRIPPERS_YAML=mycell.yaml`) is parsed as a **named volume**, so an empty directory gets mounted over `/config/grippers.yaml`;
- relative paths resolve against the compose file's directory (`docker/`), not the user's cwd.

Worth a line in the header saying it must be an absolute path, or a `./`-prefixed one relative to `docker/`.

mcp/demo/demo.sh
```diff
@@ -0,0 +1,24 @@
...
+# Bring the no-hardware demo up (driver on fake hardware + RViz + MCP server)
+# or down: ./mcp/demo/demo.sh [up|down]
+set -euo pipefail
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PORT="${GRIPPER_MCP_PORT:-8301}"
+
+case "${1:-up}" in
+  up)
+    xhost +local: >/dev/null 2>&1 || echo "xhost not available; RViz may not open"
+    docker compose -f "${HERE}/docker-compose.yml" up -d
```
mbegin-robotiq, line 24: **`demo.sh` can't build what it needs, and says so badly.**

`up -d` without `--build`, and the demo compose has no `build:` section — both services take `image:` only. If either image is missing locally, Docker tries to pull it and the user gets `pull access denied for robotiq_gripper_mcp` — which reads like an auth problem, not "run the two documented builds first".

A preflight `docker image inspect "$IMAGE" >/dev/null 2>&1 || { echo "build the base images first: …"; exit 1; }` would turn that into the actual instruction.

mcp/Dockerfile
```diff
@@ -0,0 +1,45 @@
+# gripper_mcp on top of this repo's ROS 2 image (docker/Dockerfile): that image
+# already has rclpy, control_msgs and robotiq_tsf's messages on the distro's
+# Python, so the server here only adds its own pure-Python dependencies.
+#
+# Build the base first, then this one with mcp/ as the context:
+#   docker build -f docker/Dockerfile -t robotiq_ros2:jazzy .
+#   docker build -t robotiq_gripper_mcp:jazzy mcp
+#
+# Jazzy and Lyrical only: fastmcp needs Python 3.12 and Humble ships 3.10.
```
mbegin-robotiq: **"Jazzy and Lyrical only" is only a comment.**

A Humble base builds cleanly all the way through — `uv.lock` says `>=3.10` and the wheels are `py3-none-any` — and fails only at container start, on the fastmcp import. Since `ROS_DISTRO` comes from the base image, a guard right after the `FROM` turns that into a build-time error at the line that documents the constraint:

```dockerfile
RUN case "${ROS_DISTRO}" in jazzy|lyrical) ;; *) echo "gripper_mcp needs Python 3.12: jazzy or lyrical, got ${ROS_DISTRO}" >&2; exit 1 ;; esac
```

mbegin-robotiq (approved):
Approving — the `BASE_IMAGE`-arg-over-third-stage call is the right one and well argued, and the venv-on-the-system-interpreter trick is exactly what lets one Python import both `rclpy` and `fastmcp`. I checked the ROS side: the launch args exist, the rviz config path matches the colcon install layout, `namespace: /` passes the strict config model, and the base image's `ENTRYPOINT ["/bin/bash"]` is what makes the `-c` commands work.

Six inline comments. Only the first is worth blocking on in my view — the demo's `ROS_DOMAIN_ID` default doesn't hold on a real cell, which turns a no-hardware demo into something that can command real hardware. The rest are footguns and docs.

mcp/Dockerfile
ebarnett3: ### 1. "Humble's Python 3.10 cannot run fastmcp" is false — Low (doc), and it contradicts mbegin's thread

`Dockerfile:9` says "Jazzy and Lyrical only: fastmcp needs Python 3.12 and
Humble ships 3.10", and `README.md:145` repeats it. The same PR ships
`pyproject.toml:7` with `requires-python = ">=3.10"` and `fastmcp==4.0.2`,
`ci-mcp.yml` with a `["3.10", "3.12"]` matrix, and `README.md:132-133` saying
"It runs on Python 3.10 and up, Humble's interpreter and Jazzy's, and CI tests
both".

Checked: the whole suite, `fastmcp` server tests included, passes on 3.10.

```
=== py3.12 run: 160 passed in 1.44s
=== py3.10 run: 160 passed in 1.53s
```

So whatever restricts the image to Jazzy and Lyrical, it is not fastmcp.

This matters for mbegin's thread on `Dockerfile:9`, which proposes a build-time
guard on `ROS_DISTRO` and states that a Humble base "fails only at container
start, on the fastmcp import". That premise does not hold: fastmcp 4.0.2
imports and runs on 3.10, as the passing suite shows, and `requires-python` and
the CI matrix both say so. Adding a guard would enforce a restriction that has
no cause behind it. Worth settling in that thread before anyone implements it:
either name the real constraint, or drop the claim and let a Humble base
build.

mcp/demo/demo.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+# Bring the no-hardware demo up (driver on fake hardware + RViz + MCP server)
+# or down: ./mcp/demo/demo.sh [up|down]
+set -euo pipefail
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PORT="${GRIPPER_MCP_PORT:-8301}"
+
+case "${1:-up}" in
+  up)
+    xhost +local: >/dev/null 2>&1 || echo "xhost not available; RViz may not open"
```
ebarnett3: ### 2. `demo.sh` opens the X display to every local user and never closes it — Low

`demo/demo.sh:10` runs `xhost +local:` and the `down` branch
(`demo/demo.sh:17-19`) does not revoke it, so a demo leaves the display open
until the next login. It is a demo script on a developer machine, so the stakes
are low, but the narrower grant costs nothing:

```bash
xhost +local:root >/dev/null 2>&1 || echo "xhost not available; RViz may not open"
```

and `xhost -local:root` in the `down` branch. The container runs as root, which
is what `local:root` covers.

mcp/demo/grippers.yaml
```diff
@@ -0,0 +1,10 @@
+# Wiring for the demo stack: one 2F-85 on the real driver, running
+# ros2_control's fake hardware at the root of the graph (hence namespace: /).
+# Fake hardware moves instantly and never stalls, so gripper_grasp reports
+# closed_without_object; a stall, and the tactile tools, need the real gripper
+# or the SDK's simulated one.
```
ebarnett3: ### 3. The demo's wiring file explains the outcome with the wrong reason — Low (doc)

`demo/grippers.yaml:3-5` says "Fake hardware moves instantly and never stalls,
so gripper_grasp reports closed_without_object". The conclusion is right, the
reason is not: `mock_components/GenericSystem` leaves the velocity state at 0,
which is precisely the condition that makes the controller report
`stalled: true` on every goal (#29). What actually produces
`closed_without_object` is that the fake reports exactly the commanded angle, so
the achieved opening is 0.0 mm and the service's tolerance is not crossed.

Worth getting right in this file, and more so now that jlongval has confirmed
real grippers are in scope, because it is the sentence that hides #59
finding 1: on real hardware the same call lands 0.75 mm short of the commanded
angle and reports `grasped`. A reader who believes the fake "never stalls" will
not expect that.

Suggested:

```
# Fake hardware reports exactly the angle it was commanded, so a grasp closes to
# 0 mm and reports closed_without_object. A real gripper stops short on an object
# (and reports stalled on every goal, see #29); the tactile tools need real pads.
```

docker/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    container_name: robotiq_gripper_mcp
+    network_mode: host
+    ipc: host
+    restart: unless-stopped
```
ebarnett3, line 28: ### 4. No healthcheck on a `restart: unless-stopped` service — Nit

`docker/docker-compose.yml:12-21`. The server answers HTTP on a known port, so
a healthcheck is one line and turns a crash-looping container into a visible
`unhealthy` state:

```yaml
healthcheck:
  test: ["CMD-SHELL", "curl -fsS http://127.0.0.1:${GRIPPER_MCP_PORT:-8300}/mcp -o /dev/null || exit 1"]
  interval: 30s
  timeout: 5s
  retries: 3
```

(`curl` has to be in the base image; drop the finding if it is not.)

mcp/README.md
```diff
@@ -127,6 +136,76 @@ It talks to the driver over ROS 2 topics and actions at runtime, so it needs a
...
+both `rclpy` and `fastmcp`. Jazzy and Lyrical only: Humble's Python 3.10 cannot
+run fastmcp. Build the base image first, with the repo root as its context:
+
+```bash
+docker build -f docker/Dockerfile -t robotiq_ros2:jazzy .
+docker build -t robotiq_gripper_mcp:jazzy mcp
+```
+
+`BASE_IMAGE` (default `robotiq_ros2:jazzy`) selects any image built from
+`docker/Dockerfile`: a headless build, or something layered on top of it.
```
ebarnett3: I dont think we need to repeat the comments from the Dockerfile here. We could for example just have a short sentence here pointing to the details in Dockerfile.

mcp/README.md
```diff
@@ -127,6 +136,76 @@ It talks to the driver over ROS 2 topics and actions at runtime, so it needs a
...
+with [`demo/grippers.yaml`](demo/grippers.yaml): one `driver` gripper on that
+driver, at the root of the graph (`namespace: /`).
+
+```bash
+./mcp/demo/demo.sh          # xhost + compose up; `demo.sh down` tears it down
+```
+
+Closing the RViz window ends its container; the next `demo.sh` brings it back.
+On a headless base image start only `gripper-driver gripper-mcp` with compose.
+The stack sits on `ROS_DOMAIN_ID=42` and port 8301 so it never joins a real cell's graph.
```
ebarnett3: Here we repeat a lot of the text in docker-compose.yml. Same comment as for the Dockerfile.

ebarnett3 (approved):

mcp/demo/docker-compose.yml
jlongvalRobotiq: Fixed: all three demo services now set `ROS_DOMAIN_ID: "42"` outright, and the header says it is fixed, not read from the environment. Reran the demo on a host that exports a cell domain: the MCP container reports 42.

docker/docker-compose.yml
jlongvalRobotiq: Default is `../mcp/grippers.yaml.example` now, so a fresh checkout starts with the example wiring; the header says why.

docker/docker-compose.yml
jlongvalRobotiq: `GRIPPER_MCP_IMAGE` knob on both compose files, default `robotiq_gripper_mcp:jazzy`, and the header says to keep it in step with the base. A lyrical build lands on whatever tag the caller names instead of overwriting jazzy.

docker/docker-compose.yml
jlongvalRobotiq: Header now says it is a compose bind source: absolute or `./`-prefixed, resolved against `docker/`, or compose reads it as a named volume. The example command uses `./../mcp/grippers.yaml`.

mcp/demo/demo.sh
jlongvalRobotiq: `require_image` in `demo.sh` inspects both images before `up` and prints the two build commands with the actual tags. Tried it with a missing tag: exit 1 and the two commands, no `pull access denied`.

mcp/Dockerfile
jlongvalRobotiq: Eric is right that the constraint was not real: the suite passes on 3.10 and `requires-python` says so. Dropped the claim from the Dockerfile header and the README rather than guarding it. A Humble base is allowed to build; nothing in this PR has been run on one, so the README does not promise it either.

mcp/demo/demo.sh
jlongvalRobotiq: `xhost +local:root` on `up`, `xhost -local:root` on `down`, both tolerant of a missing `xhost`. Checked the grant is gone after `down`.

mcp/demo/grippers.yaml
jlongvalRobotiq: Took your wording: the fake reports exactly the commanded angle, so the grasp closes to 0 mm and reports `closed_without_object`; a real gripper stops short and reports stalled on every goal (#29); the tactile tools need real pads.

docker/docker-compose.yml
jlongvalRobotiq: Dropped, as you allowed: `docker/Dockerfile` does not install curl and adding it there for a healthcheck is out of this PR's scope. Worth a line in a follow-up if the base image grows curl.

mcp/README.md
jlongvalRobotiq: The README now points at the Dockerfile header for the base image, venv and build commands and keeps only the three commands.

mcp/README.md
jlongvalRobotiq: Same: one sentence pointing at the compose header for the knobs and the wiring-file rules, the duplicated text is gone.

PR #73, opened 2026-09-10
feat(mcp): gripper_grasp_until_contact, tactile-guided closing

jlongvalRobotiq:
Stacked on #66. The tactile-guided closing loop split out of #66: close by the datasheet's `step_mm`, read the pads between steps, stop on first touch, on the fingers meeting, on a stall, on a refusal, or on the hard `contact_timeout_s`. A stall with quiet pads is `stalled_before_contact`, never a grasp. Pure stop rules in `contact_grasp.py`; the only tool that issues repeated motion from inside one call.

**Draft until run on a real 2F-85 with TSF-85 pads.** Unit tests: 169 passed (9 new) on the fake gripper and pads.


jlongvalRobotiq:
Parked: the contact grasp needs a real 2F-85 with TSF-85 pads to be worth reviewing, and none are on hand. The branch stays. When it comes back it should stop on the tare-derived `contact` from `TactileService.read()` (#65) and read a stopped step from the achieved opening rather than the `stalled` flag (#29, #74).

PR #78, opened 2026-09-15
Issue 74/publish gobj

### LINKED ISSUE #31: Expose the full gripper command and status blocks to ROS
The SDK exposes the full gripper command and status blocks, but the ROS layer does not pass most of it through. A ROS consumer today sees position (state + command), a derived velocity, and the two activation command interfaces — everything else the gripper reports on every cycle stops at the hardware interface.

What should be reachable from ROS:

- **Status**: the complete status block — `gACT`, `gGTO`, `gSTA`, `gOBJ`, `gFLT`, `gPR`, `gPO`, `gCU`. Object detection and current are tracked separately in #26 as `ros2_control` state interfaces; this issue is the broader question of publishing the whole block.
- **Command**: the values actually being written each cycle (`rACT`, `rGTO`, `rATR`, `rPR`, `rSP`, `rFR`), so a consumer can see what the driver is sending rather than inferring it.

## Open questions

- **Interface choice.** `ros2_control` state/command interfaces cover what a controller needs, but they are a poor fit for diagnostics — nothing outside the controller manager can read them. A latched topic (or a diagnostics topic) publishing the decoded blocks is the natural complement. Deciding which values go where is the substance of this issue.
- **Rate.** The exchange cycle runs far faster than anything needs to consume the status for display. A publisher needs its own rate, decoupled from the control loop.
- **Naming.** Whether topics expose raw register names (`gPR`, `gOBJ`) or decoded semantics. Raw names match the manual and are unambiguous; decoded names are friendlier to someone who has never read it.

An inspection/control surface for basic commands (rviz panel or a small web viewer) is the eventual goal, but it depends on this being published first.

### LINKED ISSUE #74: Publish object_status (gOBJ) and motor_current so clients can read them
The driver reads the gripper's own object-detection field (gOBJ) every cycle and exports it as a state interface: `hardware_interface.cpp:397` assigns `status.gripperStatus.objectDetection()` to `gripper_object_status_`, registered as `object_status` on the command joint (`hardware_interface.cpp:288`). It also exports `motor_current`. Neither reaches the graph: `joint_state_broadcaster` publishes only position, velocity and effort, and no other broadcaster claims them. So the one authoritative answer to "is the gripper holding something" stops inside the controller manager.

Everything above the driver therefore guesses. That matters now rather than
eventually: the MCP server is confirmed for use on real grippers, and its grasp
verdict is one of these inferences. The `gripper_cmd` action's `stalled` flag is the usual substitute, and #29 shows it is stuck true on every goal because joint velocity is never computed. The MCP server (#59) infers a grasp from the difference between the commanded and achieved opening. Both are inferences about a fact the hardware reports directly.

## Scope

- A broadcaster in `robotiq_controllers`, alongside `robotiq_activation_controller`, claiming the driver's non-standard state interfaces and publishing them on a latched-free topic at the controller rate: `object_status` (gOBJ) and `motor_current`.
- A message for it. gOBJ is a four-state enum (moving, detected while opening, detected while closing, at requested position); publishing a bare number would push the decoding onto every client, so the message should name the states as constants.
- Added to the shipped controller configs and spawned by `robotiq_control.launch.py` like the other controllers.
- Absent on the fake and topic-based sim descriptions, which do not export those interfaces, so the broadcaster has to degrade rather than refuse to activate. The SDK's simulated gripper (robotiq/grippers, filed separately) is the no-hardware path that does report gOBJ, through the real driver plugin with `use_dummy`.

## Consumers

- The MCP server drops its position-based grasp inference and reports gOBJ.
- Any application that today reads `result.stalled` and gets the wrong answer.

## Related

robotiq/grippers#38 promotes the SDK's settled predicate and its gOBJ verdict into the public API, which is what the driver would evaluate per cycle to fill this topic.

## Note on #29

This does not replace fixing #29: `reached_goal` and `stalled` should also be right. It removes the need for clients to depend on them for object detection, which is not what a stall flag is for.

ebarnett3:
Closes #74. Closes #31.

The driver reads the gripper's object-detection field every cycle, and it stops
inside the controller manager. Everything above it therefore guesses at a fact
the hardware states outright, either from the `gripper_cmd` action's `stalled`
flag, which #29 shows is stuck true on every goal, or from the difference
between the commanded and the achieved opening.

What is needed is a typed topic carrying object detection. Similar needs exist for
motor current and the fault, with every code named so that no client decodes a register.

The obstacle is that the descriptions declare those interfaces for the real
driver alone. A controller naming them fails to activate on mock and
topic-based hardware, which would mean a different controller set per mode.

So this adds a `robotiq_msgs` package with one message, and a broadcaster in
`robotiq_controllers` that claims every state interface and publishes whichever
of them it finds, on `~/status` at the controller rate. Where the gripper's
fields are absent it says so once and stays quiet, which lets the same
controller set be spawned in every mode.

#31 also wanted the fault block reachable; that landed in pull request #79. 
rATR and rARD will be handled in #81. That issue has been intentionally been
deferred, since no real need for using these parameters has been identified.

FYI @jlongvalRobotiq


ebarnett3:
reminder for this PR

l-tschreiber-a11y (approved):

grippers/robotiq_controllers/src/gripper_status.cpp
```diff
@@ -0,0 +1,120 @@
...
+      if(!readings.at(field))
+      {
+         continue;
+      }
+      const std::optional<uint8_t> value = asCode(readings.at(field));
+      if(!value)
+      {
+         return std::nullopt;
+      }
+      *code = value.value();
```
mbegin-robotiq: **Absent fault fields are published as a positive "no fault".**

When the joint exports no `gripper_fault` interface, the loop skips the field and the message keeps its default, which is `GRIPPER_FAULT_NONE` / `GRIPPER_FAULT_SEVERITY_NONE`. A subscriber cannot tell "this gripper is healthy" from "this hardware never reports faults" — and the safe reading of the two is not the same one.

It is also inconsistent within the same message: `motor_current` on line 99 uses NaN for exactly this case, so `GripperStatus` ends up with two different conventions for an absent value.

One option is a `GRIPPER_FAULT_UNKNOWN` constant as the default; another is a `has_fault_status` flag. Either is fine, but the topic-based/mock setups that do not export the fault interfaces will publish this state routinely, so it is worth picking one.

grippers/robotiq_controllers/src/gripper_status_broadcaster.cpp
```diff
@@ -0,0 +1,152 @@
...
+gripper_status::Readings GripperStatusBroadcaster::read() const
+{
+   gripper_status::Readings readings;
+   for(std::size_t field = 0; field < gripper_status::FIELD_COUNT; ++field)
+   {
+      const std::optional<std::size_t>& index = interfaces_->at(field);
+      readings.at(field) =
+         index
+            ? compat::getValue(state_interfaces_.at(index.value())).value_or(std::numeric_limits<double>::quiet_NaN())
+            : std::optional<double>{};
```
mbegin-robotiq: **A failed read drops the whole cycle, including the valid `object_detection`.**

`value_or(NaN)` collapses two different outcomes into one. `getValue()` returns `nullopt` on Jazzy+ when `get_optional()` fails because the handle's lock was contended — a transient, per-cycle condition — and that becomes NaN here, indistinguishable from a genuinely out-of-range value.

Downstream in `gripper_status.cpp`, `decode()` treats a present-but-NaN fault field as a hard error and returns `nullopt` (line 112), so the entire message is dropped. `object_detection` may have read perfectly that cycle and it is discarded with it.

Propagating the `nullopt` instead would route through the `if(!readings.at(field)) continue;` path at line 105 and publish the cycle with the fault fields left alone:

```cpp
readings.at(field) = index ? compat::getValue(state_interfaces_.at(index.value()))
                           : std::optional<double>{};
```

That does collide with the point above — it makes a contended read look like an unexported interface — so the two are best resolved together.

grippers/robotiq_controllers/src/gripper_status.cpp
```diff
@@ -0,0 +1,120 @@
...
+{
+   // OBJECT_STATUS anchors the binding because decode() cannot publish without
+   // it: every object_detection value is a real state, so none can mean unknown.
+   const auto gripper = std::find_if(interfaces.begin(), interfaces.end(), [](const InterfaceName& interface) {
+      return interface.name == kInterfaceNames.at(OBJECT_STATUS);
+   });
+   if(gripper == interfaces.end())
+   {
+      return std::nullopt;
+   }
```
mbegin-robotiq: **With two grippers on one controller_manager, only the first one is ever published.**

`find_if` takes whichever interface exports `object_status` first, and its `prefix` then filters everything else. There is no `joint`/`prefix` parameter on the controller, so two instances of this broadcaster on the same controller_manager both bind to the same gripper — and the second gripper's status is never published, silently, since the "claim all, stay quiet" design has no complaint to make here.

A `joint` parameter (defaulting to empty = current behaviour) would make the target explicit and keep single-gripper configs unchanged. If dual-gripper is out of scope for this PR, a comment here saying so would at least keep the next reader from assuming it works.

grippers/robotiq_controllers/include/robotiq_controllers/ros2_control_compat.hpp
```diff
@@ -32,22 +32,27 @@
...
+// threshold, because Jazzy's 4.x line keeps moving.
 //
 // Humble (ros2_control 2.x) exposes `void set_value(double)` and
 // `double get_value()`. From Jazzy's 4.x line on, writes report success
 // (`bool set_value(double)`) and reads can fail (`std::optional<double>
 // get_optional()`), because the handles took a lock internally.
 //
+// realtime_tools split the same way: Humble publishes over `trylock()` /
+// `msg_` / `unlockAndPublish()`, and newer ones over a single `try_publish(msg)`
+// that deprecates the lock API.
```
mbegin-robotiq: **Two small things in this comment block.**

Lines 36-37 ("Each shim keys off the shape of the operation rather than a version threshold, because Jazzy's 4.x line keeps moving") restate the paragraph that is still sitting at lines 48-50 a dozen lines below. One of the two can go.

Lines 44-46 are not accurate for Humble: `realtime_tools` there already has `tryPublish()`, and it does not deprecate the lock API. The shim itself is right — it dispatches on `try_publish`, which is the member that actually differs — it is just the comment's description of the split that is off. Something like "newer releases add `try_publish(msg)`, which takes the message in one call" would match what the code does.

mbegin-robotiq (approved):
Approving — the design holds together and CI is green on humble, jazzy, lyrical and rolling. I checked the `GRIPPER_FAULT_*`, `GRIPPER_FAULT_SEVERITY_*` and `object_detection` constants against the SDK enums in `fault_status.hpp` and they all match, the interface names line up with `robotiq_driver`, `on_deactivate` resets the bound indices before `release_interfaces()`, and `robotiq_msgs` 1.1.0 is listed in both `VERSION` and `dev/CMakeLists.txt`.

Four inline notes below. The two absent-value ones (defaults reported as "no fault", and a failed read dropping a whole cycle) are the ones I would want settled; the multi-gripper binding is worth at least a decision on record, and the last one is cosmetic.

grippers/robotiq_controllers/src/gripper_status.cpp
ebarnett3: Done, with the constant rather than the flag. `GRIPPER_FAULT_UNKNOWN` and `GRIPPER_FAULT_SEVERITY_UNKNOWN` are 255, outside the four-bit register, and they are the field defaults so an unfilled message is unknown rather than healthy. It keeps the out-of-band convention `motor_current` already uses with NaN, and needs no second field to stay in sync with the first.

grippers/robotiq_controllers/src/gripper_status_broadcaster.cpp
ebarnett3: Done, as suggested.

grippers/robotiq_controllers/src/gripper_status.cpp
ebarnett3: Done, as the comment rather than the parameter. `bindInterfaces` now states that the first joint exporting `object_status` wins and that two grippers on one controller manager are unsupported. No consumer has that arrangement today, and a `joint` parameter with no user to shape it would be a guess at the interface.

grippers/robotiq_controllers/include/robotiq_controllers/ros2_control_compat.hpp
ebarnett3: Done, and the correction went further than the comment. You are right that Humble has `tryPublish` and does not deprecate the lock API; I confirmed both in a `ros:humble-ros-base` container. Since Humble offers the one-call form too, the shim's fallback now calls `tryPublish` instead of doing `trylock` / `msg_` / `unlockAndPublish` by hand, and the test's old-API fake models that. Six lines become one, and the split is a rename rather than two different shapes.

PR #79, opened 2026-09-15
feat(robotiq_driver): expose the fault block as decoded state interfaces

ebarnett3:
- add gripper_fault and fault_severity state interfaces
- decode gFLT and its severity through the SDK, so no consumer decodes the
  byte
- name the fault in the driver's warning instead of printing it as hex
- declare both interfaces on the real-hardware branch of the 2F xacros
- README: six state interfaces, gPO is a joint angle, rFR is a threshold

- update the grippers SDK module in order to use the `toString` functions


grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -76,40 +80,38 @@ constexpr auto kMotionTimeout = std::chrono::seconds{5};
...
-      return "Disconnected";
-   case Robotiq::ConnectionState::Connecting:
-      return "Connecting";
-   case Robotiq::ConnectionState::Operational:
-      return "Operational";
-   case Robotiq::ConnectionState::Faulted:
-      return "Faulted";
-   }
-   return "Unknown";
+   return std::string{name};
```
l-tschreiber-a11y: a this point could we remove function printable and just call std::string{name}  ?

grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -462,16 +474,17 @@ hardware_interface::return_type RobotiqGripperHardwareInterface::read(const rclc
                            kDiagnosticThrottleMs,
                            "The Robotiq gripper on %s: link is %s; the reported position may be stale.",
                            parameters_.connection.serial.port.c_str(),
-                           toString(connection));
+                           printable(Robotiq::toString(connection)).c_str());
```
l-tschreiber-a11y: to string -> c_str -> back to std::string ?  Are all the conversions required?

grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -462,16 +474,17 @@ hardware_interface::return_type RobotiqGripperHardwareInterface::read(const rclc
...
    }
 
    if(status.faultStatus.gripperFault() != Robotiq::GripperFault::None)
    {
       RCLCPP_WARN_THROTTLE(kLogger,
                            diagnostic_clock_,
                            kDiagnosticThrottleMs,
-                           "The Robotiq gripper on %s reports fault status 0x%02X.",
+                           "The Robotiq gripper on %s reports fault %s (gFLT 0x%02X).",
```
l-tschreiber-a11y: I assume gFLT stands for gripperFault. We should avoid abbreviations. I recommend using gripperFault instead.

README.md
```diff
@@ -338,22 +338,32 @@ ros2 action send_goal /robotiq_gripper_controller/gripper_cmd \
 
 `position` is the joint angle in radians (≈ `0.0` open → ~`0.8` closed on a 2F-85); on Jazzy/Lyrical `effort` and `velocity` are optional max limits, mapped to the controller's `set_gripper_max_effort` / `set_gripper_max_velocity` interfaces.
 
+`effort` sets the gripper's grip threshold, as a fraction of `gripper_max_force` written to rFR. It is what the fingers push with once they meet an object, not a force the gripper regulates to or reports back — nothing reads a force out of the gripper, and `motor_current` is not convertible to one.
```
l-tschreiber-a11y: effort sets the gripper's grip threshold as a fraction of gripper_max_force written to rFR. It sets the maximum current at the motor but DOES NOT directly control the max force applied by the gripper, as the max force is also largely influenced by the closing speed — nothing reads a force out of the gripper, and motor_current is not convertible to one.

grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -177,6 +179,8 @@ hardware_interface::CallbackReturn RobotiqGripperHardwareInterface::on_init(cons
    gripper_velocity_ = std::numeric_limits<double>::quiet_NaN();
    gripper_motor_current_ = std::numeric_limits<double>::quiet_NaN();
    gripper_object_status_ = std::numeric_limits<double>::quiet_NaN();
+   gripper_fault_ = std::numeric_limits<double>::quiet_NaN();
+   fault_severity_ = std::numeric_limits<double>::quiet_NaN();
```
l-tschreiber-a11y: I see those two params grouped together everywhere. Could we group them in a struct so they are passed as a single thing: 
```
struct GripperFault
{
    int  type 
    int  severity
}

**Review by mbegin-robotiq — COMMENTED** [review]
Three points from a full pass over the diff, including the `extern/grippers` bump. The `printable` question already open in the other thread still stands separately.

File: grippers/robotiq_driver/src/hardware_interface.cpp [hunk 6]
```diff
@@ -462,16 +474,17 @@ hardware_interface::return_type RobotiqGripperHardwareInterface::read(const rclc
...
    if(status.faultStatus.gripperFault() != Robotiq::GripperFault::None)
    {
       RCLCPP_WARN_THROTTLE(kLogger,
                            diagnostic_clock_,
                            kDiagnosticThrottleMs,
-                           "The Robotiq gripper on %s reports fault status 0x%02X.",
+                           "The Robotiq gripper on %s reports fault %s (gFLT 0x%02X).",
                            parameters_.connection.serial.port.c_str(),
+                           printable(Robotiq::toString(status.faultStatus.gripperFault())).c_str(),
                            status.faultStatus.raw());
```
mbegin-robotiq: The format now says `gFLT`, but `raw()` is the **whole** FAULT STATUS byte, kFLT included.

From the SDK this PR bumps to: `detail/register_map.hpp` defines `kControllerFaultMask = 0xF0` (kFLT, high nibble), and `test_typed_layers.cpp` asserts `raw() == 0x5C` decodes to `kFLT 0x5` / `gFLT 0xC`.

So with gFLT = OverTemperature and any controller fault latched, this logs `fault OverTemperature (gFLT 0xC8)` — `0xC8` is not a gFLT code at all. It also contradicts the README added in this same PR, which advertises the format as `reports fault OverTemperature (gFLT 0x08)`.

The old message (`"reports fault status 0x%02X"`) was correct with `raw()`; relabelling it as gFLT is what broke it. Suggest:

```suggestion
                           static_cast<unsigned>(status.faultStatus.gripperFault()));
```

grippers/robotiq_driver/tests/test_robotiq_gripper_hardware_interface.cpp
```diff
@@ -266,6 +274,36 @@ TEST(TestRobotiqGripperHardwareInterface, UseDummyActivatesAndFollowsCommands)
...
+                                  hardware_interface::lifecycle_state_names::ACTIVE};
+   ASSERT_EQ(hardware_interface::return_type::OK, rm.set_component_state(kComponentName, active));
+
+   auto gripper_fault = rm.claim_state_interface("robotiq_85_left_knuckle_joint/gripper_fault");
+   auto severity = rm.claim_state_interface("robotiq_85_left_knuckle_joint/fault_severity");
+
+   ASSERT_TRUE(compat::readWriteOk(rm.read(rclcpp::Time{0}, rclcpp::Duration::from_seconds(0.01))));
+
+   EXPECT_DOUBLE_EQ(static_cast<double>(Robotiq::GripperFault::None), compat::getValue(gripper_fault).value_or(-1.0));
+   EXPECT_DOUBLE_EQ(static_cast<double>(Robotiq::FaultSeverity::None), compat::getValue(severity).value_or(-1.0));
```
mbegin-robotiq, line 305: Both expected values are zero — `GripperFault::None == 0` and `FaultSeverity::None == 0` — so this test cannot fail for the bug class its docstring names. It passes just as happily if the driver hardcodes `0.0`, if the two fields are swapped, or if `fault_severity_` were assigned `gripperFault()` instead of `severity(...)`.

The docstring claims it shows the interfaces "carry the SDK's decoded code and severity, not the raw byte"; as written it shows neither.

---

**Correcting my own suggestion above.** I originally proposed asserting on a faulted gripper via the SDK's fake. That is not possible, and the reason matters: `fake/gripper_factory.hpp` states the fake device has "no motion profile, no travel time, no object detection and **no fault injection**." The injection used in the SDK's own `test_fake_gripper_server.cpp` comes from `#include "fake/status_writer.hpp"`, a relative include into `src/fake/` that is not among the installed headers. So this package cannot reach it.

Under `use_dummy`, `gripper_fault` is therefore structurally pinned to `None`, and no test written in this shape can ever cover the decoding.

The way out is the pattern already in this package: `motorCurrentFromRegister` in `gripper_scaling.hpp` is a pure inline function, table-tested in `test_gripper_scaling.cpp` with no gripper involved. Pulling the fault decode into the same header — something like `faultCodeFrom(FaultStatus)` and `faultSeverityFrom(FaultStatus)` — makes all 16 nibble values assertable directly, and one such test would also pin the `gFLT`/`raw()` mismatch and the `Unrecognized` codes raised in my other two comments.

Keeping this test as a smoke check on the unfaulted path is fine; it just should not be the thing standing in for coverage of the decoding.

README.md
```diff
@@ -338,22 +338,32 @@ ros2 action send_goal /robotiq_gripper_controller/gripper_cmd \
...
 
 | Interface | Unit | Description |
 |---|---|---|
-| `position` | rad | Knuckle joint angle, from gPO (≈ `0.0` open → ~`0.8` closed on a 2F-85) |
+| `position` | rad | Angle of the driven knuckle joint, from gPO — not the opening in millimetres. `0.0` open → ~`0.8` closed on a 2F-85, over a linear approximation of a travel the gripper reports in 227 counts |
 | `velocity` | rad/s | Always `0.0`. The status block carries no velocity, and the driver does not differentiate the position |
 | `motor_current` | A | gCU, the motor current the manual gives as 10 mA per count, so `0.0` to `2.55` |
 | `object_status` | — | gOBJ verbatim: `0` moving, `1` object held while opening, `2` object held while closing, `3` at the requested position |
+| `gripper_fault` | — | gFLT as the manual numbers it: `0` none, `5` action delayed, `7` activation required, `8` over temperature, `9` no communication, `10` under voltage, `11` automatic release running, `12` internal fault, `13` activation fault, `14` overcurrent, `15` automatic release complete |
```
mbegin-robotiq: This table reads as the contract, but `gripperFault()` returns the masked low nibble, so `1`, `2`, `3`, `4` and `6` are all reachable on the wire and have no enumerator in the SDK. It prints them as `Unrecognized(0x6)`, and `severity()` deliberately maps them to `Major` so an unknown fault is never mistaken for harmless.

Since a consumer reading `gripper_fault` can see those values, worth a closing clause that any code not listed here is an undefined fault and reports severity `3`. Otherwise someone writes a switch over the eleven documented codes and silently drops the ones that matter most.

mbegin-robotiq (approved):

grippers/robotiq_driver/src/hardware_interface.cpp
```diff
@@ -76,40 +80,38 @@ constexpr auto kMotionTimeout = std::chrono::seconds{5};
...
-                      "Joint '%s' has %s state interface. Expected %s, %s, %s or %s.",
+                      "Joint '%s' has %s state interface. Expected one of %s.",
                       joint.name.c_str(),
                       state_interface.name.c_str(),
-                      hardware_interface::HW_IF_POSITION,
-                      hardware_interface::HW_IF_VELOCITY,
-                      kMotorCurrentInterface,
-                      kObjectStatusInterface);
+                      expected.str().c_str());
          return false;
```
mbegin-robotiq, line 107: This validator is rewritten here — a chain of `||` becomes `std::find` over a `std::array`, and the error message is rebuilt from the same array — and nothing in the suite exercises it.

I grepped `robotiq_driver/tests/`: no test feeds `on_init` a URDF declaring an unsupported state interface, so the `return false` path has never run, before this PR or after. The rewrite itself looks correct to me (`std::find` compares `const char*` against `state_interface.name` through `operator==(const char*, const std::string&)`, so it is a content comparison, and the `ostringstream` is only built on the failure path) — but a validation rewrite with zero coverage is the riskiest thing in this diff, and the blast radius is a gripper that silently accepts a description it should reject.

A test adding `<state_interface name="effort"/>` to `minimalRobotUrdf()` and asserting the component fails to initialise would cover it, and would have to be updated deliberately the next time this list grows — which is the point.

Two smaller gaps while this is open, both pre-existing rather than introduced here, but both widened by the two new fields:

- The README's "All six read `NaN` until the component is activated" is asserted nowhere. Worth one test reading the interfaces before `on_activate`, since the sentinel is load-bearing for `object_status` and now for `gripper_fault` too, where `0` is a real value.
- The decoding gap in my comment on the new test, which the extraction suggested there would close.

mbegin-robotiq (commented):
One more on test coverage, after checking what the suite currently exercises.

grippers/robotiq_driver/src/hardware_interface.cpp
ebarnett3: done

grippers/robotiq_driver/src/hardware_interface.cpp
ebarnett3: The multiple conversions have been eliminated. Now we construct a single `std::string` and then use `c_str()`.

grippers/robotiq_driver/src/hardware_interface.cpp
ebarnett3: changed to "gripper fault". In general I agree that using complete words is preferable. However, I think there are a lot of users who are quite familiar with the gripper register names.

README.md
ebarnett3: Done, as written. Also applied to the `gripper_max_force` row, which was wrong the same way.

grippers/robotiq_driver/src/hardware_interface.cpp
ebarnett3: Done, as `FaultState` with `double` members. `GripperFault` would collide with the SDK's type of that name, and a `StateInterface` binds by address and holds only `double`.

grippers/robotiq_driver/src/hardware_interface.cpp
ebarnett3: Done.

grippers/robotiq_driver/tests/test_robotiq_gripper_hardware_interface.cpp
ebarnett3: test added to `test_gripper_scaling.cpp`: verifies that a raw fault register with gripper and controller faults is decoded correctly.

README.md
ebarnett3: Done, in the README table and on the message constants.

grippers/robotiq_driver/src/hardware_interface.cpp
ebarnett3: Done, but not through `ResourceManager`: that route passes with the validator compiled out, because ros2_control rejects such a description first. The test drives `on_init` directly, with a companion asserting success for a supported name.

PR #83, opened 2026-09-16
mcp: tare over 100 samples, about 1 s

jlongvalRobotiq:
Follow-up to the #66 review, tare thread: https://github.com/robotiq/ros/pull/66#discussion_r3993798238

## Why

`baseline_samples: 1000` in the TSF-85 datasheet was copied from the SDK's `findBaseline()`. On #66 the tare became `sample(count)` over distinct frames, and the jazzy rerun showed what that count costs: rclpy delivers about 110 `StaticData` callbacks a second to Python whatever the pads publish (~2 kHz on Eric's bench), so `gripper_tare_tactile` took 8.8 s. mbegin had proposed 60 frames over 1 s; Eric closed the thread with "I agree that 9 s is too long. 1 s seems like a good compromise." This PR is that compromise.

## What

- `robotiq_tsf_85.yaml`: `baseline_samples: 100`, comment says why the SDK's 1000 is not the right number here.
- `test_config.py`: the `gt=0` validation test edits the new value.

Nothing else changes: the noise floor and the threshold are still measured over whatever count the datasheet gives. With a rest noise of +-2 counts, 100 frames fix the baseline to a fraction of a count.

## Verified

- `uv run pytest`: 189 passed.
- pre-commit clean on the two files.
- Not rerun live: the 8.8 s figure for 1000 frames is from the #66 body; 100 frames at the same callback rate is about 0.9 s.


mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
```diff
@@ -28,5 +28,7 @@ contact_threshold: 0.02
 # hold reads about 0.8, so 2x the noise separates the two with room.
 noise_margin: 2.0
 
-# Matches the tactile SDK's findBaseline(), which averages 1000 samples.
-baseline_samples: 1000
+# Distinct frames averaged by a tare. The SDK's findBaseline() takes 1000, but
+# rclpy hands Python about 110 frames a second whatever the pads publish, so
```
ebarnett3, line 32: Is this throttling to 110 frames per second something we chose?

ebarnett3 (approved):

mcp/gripper_mcp/datasheets/tactile/robotiq_tsf_85.yaml
jlongvalRobotiq: No, nothing sets it. It's what we measured in the #66 live run on my workstation (jazzy container, dummy publisher at ~1.7 kHz): the 1000-frame tare took 8.8 s, so about 110 frames a second reach Python. 100 frames is sized so the tare takes about 1 s at that rate. Reworded the yaml comment to say so (09199b3).

PR #86, opened 2026-09-16
mcp: refuse a stale joint state instead of serving it as current

jlongvalRobotiq:
Third of three fix PRs from the acceptance matrix run (row 12). Stacked on #85.

## Why

`gripper_get_state` waited for any `joint_states` message ever received and returned it with `measured_at` set to now. With the driver container gone for 25 s it still answered `opening_mm: 85.0` as a fresh reading. `gripper_get_health` already treated a state older than 2 s as stale; the two disagreed.

## What

- `ros_backend.py`: `JointStateFeed` gains `age_s()` and `fresh()`; `_await_state` polls for a message younger than `STALE_STATE_S` instead of any message. `read_state` and `health` share that one definition of fresh, so the age check that lived only in `health` is gone.
- `ros_messages.py`: `no_state_message` builds the error, naming the age of the last message when there was one, or the wait when there was none. `JOINT_STATES_TOPIC` moves here so the message and the subscription use the same name.
- `test_ros_messages.py`: both wordings.

## Verified

- `uv run pytest`: 195 passed.
- Live on the demo: with the driver stopped, `gripper_get_state` -> "The last joint_states naming 'robotiq_85_left_knuckle_joint' under '/' is 13.1 s old (limit 2 s); has the controller stopped?", again at 34.7 s; health says `joint_states stale`; after `docker start` on the driver, state reads normally again.


mcp/gripper_mcp/ros_backend.py
```diff
@@ -185,9 +195,12 @@ def read_state(self) -> BackendState:
...
         if not self._await_state():
             raise RuntimeError(
-                f"No {JOINT_STATES_TOPIC} naming '{joint}' under "
-                f"'{self._node.get_namespace()}' within {STATE_WAIT_S:.0f} s; "
-                "is the controller running?"
+                no_state_message(
+                    self._states.age_s(),
+                    joint,
+                    self._node.get_namespace(),
+                    STALE_STATE_S,
```
mbegin-robotiq: Nit: when no message ever arrived, `no_state_message` renders this as "within X s", which describes how long we waited, i.e. `STATE_WAIT_S`, not the staleness limit. Both are 2.0 today so the output is right by coincidence; if `STATE_WAIT_S` is raised to 5 the call waits 5 s but still says "within 2 s". Consider passing both values (wait for the `None` case, limit for the stale case).

mcp/gripper_mcp/ros_backend.py
```diff
@@ -185,9 +195,12 @@ def read_state(self) -> BackendState:
         joint = self.joint_geometry().name
         if not self._await_state():
             raise RuntimeError(
-                f"No {JOINT_STATES_TOPIC} naming '{joint}' under "
-                f"'{self._node.get_namespace()}' within {STATE_WAIT_S:.0f} s; "
-                "is the controller running?"
+                no_state_message(
+                    self._states.age_s(),
```
mbegin-robotiq: Nit: the age is re-read after `_await_state()` has given up, so it can disagree with the check that just failed. `_on_joint_states` also sets `latest` before `latest_at` (no lock), which gives two windows:
- first message lands as the wait ends → `latest` set, `latest_at` still 0.0 → error reports an age equal to the machine uptime (e.g. "183422.7 s old").
- a message arrives just after the wait fails → "0.0 s old … has the controller stopped?" about a state that is actually fresh.

Assigning `latest_at` before `latest` in the callback fixes the first; capturing the age inside the poll (or re-checking freshness here) fixes the second.

mbegin-robotiq (approved):
LGTM — `gripper_get_state` and `gripper_get_health` now share one freshness check, so a stale reading is no longer served as current. Two minor, non-blocking nits inline about the error text.

